PR #5, opened 2021-11-04
Add test client binary

mrunalp:
@haircommander @saschagrunert ptal

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>


saschagrunert (approved):
A client is definitely something we should add!

PR #6, opened 2021-11-05
Bump dependencies and remove glib

saschagrunert:
I think we wanted to get rid of the glib dependency, so we can remove it
from the crate as well. We now also bump the dependencies by using their
exact version number, they're still locked via Cargo.lock.


haircommander:
LGTM

PR #8, opened 2021-11-08
WIP: Switch to capnproto

saschagrunert:
Just an experiment using the cap'n proto RPC framework:

```
> cat /proc/(pidof conmon-server)/status | rg RSS
VmRSS:      2260 kB
```

Compared to `main`:
```
> cat /proc/(pidof conmon-server)/status | rg RSS
VmRSS:      3312 kB
```

PR #9, opened 2021-11-08
Switch to single threaded tokio runtime

saschagrunert:
This saves more than 50% RSS memory on runtime.

Before:
```
VmRSS:      3264 kB
```

After:
```
VmRSS:      1460 kB
```

Way better than: https://github.com/containers/conmon-rs/pull/8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -440,16 +440,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "num_cpus"
-version = "1.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


haircommander:
Eventually, I think we will want some pieces of conmon multithreaded--each container in the pod should not wait on the others for processing container stdout/catching exits. Though, we can build on top of the single tokio's if tokio's is so heavy

rphillips:
@haircommander this change allows for a single threaded event loop, but it should allow for the usual async processing of container events.

haircommander:
ah I see, thanks!

haircommander:
LGTM

PR #10, opened 2021-11-08
Remove unused dependencies

saschagrunert:
We now do not rely on dependencies any more which we do not actually
use.


haircommander:
LGTM

PR #11, opened 2021-11-08
Validate config before server start

saschagrunert:
Before applying this patch the function was basically unused.


haircommander:
LGTM

PR #15, opened 2021-11-09
Add max rss integration test

saschagrunert:
This adds a test to verify the maximum amount of RSS used by the server binary.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1,0 +1,65 @@
+use anyhow::{Context, Result};
+use nix::{
+    sys::signal::{kill, SIGINT},
+    unistd::Pid,
+};
+use tokio::{
+    fs,
+    process::Command,
+    time::{sleep, Duration},
+};
+
+// The maximum amount of allowed VmRSS in Kilobytes
+const MAX_RSS_KB: u32 = 4000;
+
+// We assume that the tests run in release mode
+const SERVER_BINARY: &str = "target/release/conmon-server";
+const CLIENT_BINARY: &str = "target/release/conmon-client";
+
+#[tokio::test]
+async fn rss_verification() -> Result<()> {
+    // Start the server
+    let mut server = Command::new(SERVER_BINARY).arg("--runtime=/tmp").spawn()?;
+    let pid = server.id().context("no pid for child")?;
+    tokio::spawn(async move {
+        server.wait().await.expect("wait for server process");
+    });
+
+    // Wait for the server up and running
+    for i in 1..101 {
+        let status = Command::new(CLIENT_BINARY).status().await?;
+        if status.success() {
+            break;
+        }
+        assert!(i != 100);
+        sleep(Duration::from_millis(200)).await;
+    }
+
+    // Retrieve the RSS
+    let contents = fs::read_to_string(format!("/proc/{}/status", pid)).await?;
+    let mut rss_res = "";
+    for line in contents.lines() {
+        if line.starts_with("VmRSS:") {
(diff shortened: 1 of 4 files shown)


tests/integration.rs
```diff
@@ -0,0 +1,55 @@
...
+        let status = Command::new(CLIENT_BINARY).status()?;
+        if status.success() {
+            break;
+        }
+        assert!(i != 100);
+        thread::sleep(Duration::from_millis(200));
+    }
+
+    // Retrieve the RSS
+    let contents = fs::read_to_string(format!("/proc/{}/status", pid))?;
```
saschagrunert, line 35: Is this something we can trust?

tests/integration.rs
```diff
@@ -0,0 +1,55 @@
...
+use std::{fs, process::Command, thread, time::Duration};
+
+// The maximum amount of allowed VmRSS in Kilobytes
+const MAX_RSS_KB: u32 = 4000;
+
+// We assume that the tests run in release mode
+const SERVER_BINARY: &str = "target/release/conmon-server";
+const CLIENT_BINARY: &str = "target/release/conmon-client";
+
+#[test]
```
rphillips, line 15: if we migrate to #[tokio::test] we can leverage more of tokio by using tokio::process::Command.

saschagrunert: I’ll change it.

tests/integration.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    // Retrieve the RSS
+    let contents = fs::read_to_string(format!("/proc/{}/status", pid)).await?;
+    let mut rss_res = "";
+    for line in contents.lines() {
+        if line.starts_with("VmRSS:") {
+            rss_res = line
+                .split_whitespace()
+                .nth(1)
+                .context("split by whitespace")?;
```
rphillips, line 46: we can break in this if statement

rphillips (approved):

rphillips (changes requested):
forgot the break statement.

tests/integration.rs
```diff
@@ -0,0 +1,55 @@
...
+        }
+        assert!(i != 100);
+        thread::sleep(Duration::from_millis(200));
+    }
+
+    // Retrieve the RSS
+    let contents = fs::read_to_string(format!("/proc/{}/status", pid))?;
+    let mut rss_res = "";
+    for line in contents.lines() {
+        if line.starts_with("VmRSS:") {
```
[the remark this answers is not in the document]
haircommander: are we compiling in CI with debug or release?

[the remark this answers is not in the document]
haircommander: > which also seems to take the storage from the linked libraries into account.

I think VmRSS is a better indicator of what we're worried about. the linked libraries should all be shared in memory, right? the constant value isn't as worrying as the amount used linearly

tests/integration.rs
saschagrunert: Right, added the break.

tests/integration.rs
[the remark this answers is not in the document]
saschagrunert: > are we compiling in CI with debug or release?

We run it in release mode.

> I think VmRSS is a better indicator of what we're worried about. the linked libraries should all be shared in memory, right? the constant value isn't as worrying as the amount used linearly

Good, i just added a pmap output to the CI as additional information. Still, I'm wondering why the amount of rss literally doubles in the CI.

rphillips (approved):

PR #16, opened 2021-11-10
Add unix domain socket support

saschagrunert:
This changes the config to point to a UDS instead of a TCP listener.


saschagrunert:
Removing the WIP, the rss:
```
VmRSS:      1496 kB
```

rphillips (approved):

PR #18, opened 2021-11-10
setup oom and locale

haircommander:
```
 $ cat /proc/$(pidof conmon-server)/status | grep RSS                                                                                             
VmRSS:      1692 kB
```

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -1,0 +1,24 @@
+use anyhow::{bail, Result};
+use gettextrs::{setlocale, LocaleCategory};
+use log::info;
+use std::fs::File;
+use std::io::{ErrorKind, Write};
+
+/// Unset the locale for the current process.
+pub fn unset_locale() {
+    setlocale(LocaleCategory::LcAll, "");
+}
+
+/// Helper to adjust the OOM score of the currently running process.
+pub fn set_oom(score: &str) -> Result<()> {
+    // Attempt adjustment with best-effort.
+    if let Err(err) = File::create("/proc/self/oom_score_adj")?.write_all(score.as_bytes()) {
+        match err.kind() {
+            ErrorKind::PermissionDenied => {
+                info!("Missing sufficient privileges to adjust OOM score")
+            }
+            _ => bail!("adjusting OOM score {}", err),
+        }
+    }
+    Ok(())
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,6 +159,7 @@ dependencies = [
  "clap",
  "futures",
  "getset",
+ "gettext-rs",
  "log",
  "nix",
  "prost",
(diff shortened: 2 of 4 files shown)


src/init.rs
```diff
@@ -0,0 +1,19 @@
...
+
+/// Unset the locale for the current process.
+pub fn unset_locale() {
+    setlocale(LocaleCategory::LcAll, "");
+}
+
+/// Helper to adjust the OOM score of the currently running process.
+pub fn set_oom(score: &str) -> Result<()> {
+    // Attempt adjustment with best-effort.
+    if let Err(err)  = File::create("/proc/self/oom_score_adj")?.write_all(score.as_bytes()) {
```
rphillips, line 15: nit, If an async function calls low level primitives then we should always try the tokio::async primitive with an await. Thus a recommendation to change this function to be async. wdyt of the general practice? I worry about mixing async/non-async code, when we could always add async code to begin with.

haircommander: personally I'm more of a fan of doing it synchronously here, as we want to do this operation before proceeding with the program execution. from my perspecitve, the advantage of async would be lost because we'd immediately await anyway, thus basically making it synchronous (i think). I'm open to to discussing though

rphillips: That makes sense as well. Not a blocker to this PR.

rphillips:
@haircommander the rustfmt will need fixed

rphillips (approved):

src/init.rs
```diff
@@ -0,0 +1,19 @@
...
+/// Unset the locale for the current process.
+pub fn unset_locale() {
+    setlocale(LocaleCategory::LcAll, "");
+}
+
+/// Helper to adjust the OOM score of the currently running process.
+pub fn set_oom(score: &str) -> Result<()> {
+    // Attempt adjustment with best-effort.
+    if let Err(err) = File::create("/proc/self/oom_score_adj")?.write_all(score.as_bytes()) {
+        info!("Failed to adjust oom score {}", err)
```
saschagrunert, line 16: This function never returns an error, is this intentional?

haircommander: Yes, though in answering I've reconsidered. We need to allow permission denied to allow rootless conmon-rs, but we can error on the others

src/init.rs
```diff
@@ -0,0 +1,24 @@
...
+
+/// Helper to adjust the OOM score of the currently running process.
+pub fn set_oom(score: &str) -> Result<()> {
+    // Attempt adjustment with best-effort.
+    if let Err(err) = File::create("/proc/self/oom_score_adj")?.write_all(score.as_bytes()) {
+        match err.kind() {
+            ErrorKind::PermissionDenied => {
+                info!("Missing sufficient privileges to adjust OOM score")
+            }
+            _ => panic!("Fatal error when adjusting OOM score {}", err)
```
saschagrunert, line 20: Then we can return the error here
```suggestion
            _ => bail!("adjusting OOM score {}", err)
```

Issue #19, opened 2021-11-10
Memory usage after first tonic request


rphillips:
I was adding a RPC request to return RSS memory and noticed the RSS usage goes up in the server after the first GRPC request... significantly.

Terminal 1:
```
# target/release/conmon-server --runtime /bin/ls
```

Terminal 2:
```
# cat /proc/$(pidof conmon-server)/status|grep RSS
VmRSS:      1084 kB
# target/release/conmon-client
Version: Response { metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Wed, 10 Nov 2021 21:55:01 GMT", "grpc-status": "0"} }, message: VersionResponse { version: "0.1.0" }, extensions: Extensions }
# cat /proc/$(pidof conmon-server)/status|grep RSS
VmRSS:      3464 kB
```

saschagrunert:
@rphillips that's the reason why we have the mismatch between CI and our local tests. We test that the binary is up and running by verifying that the client works. Great find!

rphillips:
I'm still seeing this issue with #20.

saschagrunert:
> I'm still seeing this issue with #20.

I'm not on my local machine but I can reproduce in https://github.com/containers/conmon-rs/pull/22

rphillips:
In #27 we migrated to capnproto

PR #20, opened 2021-11-11
Switch to simple-logger

saschagrunert:
This should reduce the RSS memory usage and keep it constant over time
while logging.

Fixes https://github.com/containers/conmon-rs/issues/19

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -10,7 +10,7 @@ use tokio::{
 };
 
 // The maximum amount of allowed VmRSS in Kilobytes
-const MAX_RSS_KB: u32 = 4000;
+const MAX_RSS_KB: u32 = 3500;
 
 // We assume that the tests run in release mode
 const SERVER_BINARY: &str = "target/release/conmon-server";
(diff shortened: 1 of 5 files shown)


saschagrunert:
Way better than before:
```
VmRSS:      1536 kB
> ./target/release/conmon-client
Version: Response { metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Thu, 11 Nov 2021 09:33:24 GMT", "grpc-status": "0"} }, message: VersionResponse { version: "0.1.0" }, extensions: Extensions }
VmRSS:      1668 kB
```

rphillips:
It might be worth exploring the tracing crate.

rphillips:
Hmm. Tracing seems to use more memory.

tests/integration.rs
```diff
@@ -10,7 +10,7 @@ use tokio::{
 };
 
 // The maximum amount of allowed VmRSS in Kilobytes
-const MAX_RSS_KB: u32 = 4000;
+const MAX_RSS_KB: u32 = 3500;
```
haircommander, line 13: ooh the bar is raised

haircommander:
nice find LGTM

PR #21, opened 2021-11-11
Listen to SIGINT as well

saschagrunert:
If the binary runs in the terminal, then we would like to allow graceful
shutdown in the same way when receiving SIGTERM.


haircommander:
LGTM

PR #22, opened 2021-11-11
WIP: Run integration test in VM

saschagrunert:
This should give us more predictable test results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,8 @@ jobs:
       #
 
   test-integration:
-    runs-on: ubuntu-latest
+    runs-on: macos-10.15
+    timeout-minutes: 90
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


cgwalters:
FWIW in the coreos group we use two things:

- [OpenShift Prow which I've configured to support nested virt (in GCE) on build02](https://github.com/coreos/fedora-coreos-tracker/blob/main/docs/ci-and-builds.md#openshift-prow)
- CentOS CI (an OCP cluster on bare metal) where we use Jenkins

It wouldn't be really hard to onboard new repos into the GCE nested virt setup, which is a *lot* better than full qemu emulation.

Looks like [this issue](https://github.com/actions/virtual-environments/issues/183) tracks GH Actions (Azure under the covers) support.

saschagrunert:
> FWIW in the coreos group we use two things:
> 
> * [OpenShift Prow which I've configured to support nested virt (in GCE) on build02](https://github.com/coreos/fedora-coreos-tracker/blob/main/docs/ci-and-builds.md#openshift-prow)
> * CentOS CI (an OCP cluster on bare metal) where we use Jenkins
> 
> It wouldn't be really hard to onboard new repos into the GCE nested virt setup, which is a _lot_ better than full qemu emulation.
> 
> Looks like [this issue](https://github.com/actions/virtual-environments/issues/183) tracks GH Actions (Azure under the covers) support.

Thanks for the input, let's consider this and I'll defer the discussion to an issue.

PR #28, opened 2021-11-12
server: daemonize and write pidfile

haircommander:
A key aspect of conmon is that it runs independent of the parent that calls it. As such, we need to daemonize. To be able to be tracked by the parent, we also need to write the new child's pid to a file that the parent specifies, so it can be watched.

Interestingly, the RSS is greatly reduced when we do this:
```
 $ ./target/release/conmon-server --runtime /usr/bin/crun --conmon-pidfile=/tmp/pidfile
2021-11-12 14:41:28,540 INFO [conmon_server] Set log level to INFO
2021-11-12 14:41:28,540 INFO [conmon_server::init] Missing sufficient privileges to adjust OOM score
[pehunt@ovpn-114-123 conmon-rs(fork)]
 $ cat /proc/$(pidof conmon-server)/status | grep RSS
VmRSS:       248 kB
[pehunt@ovpn-114-123 conmon-rs(fork)]
 $ ./target/release/conmon-client --runtime /usr/bin/crun --conmon-pidfile=/tmp/pidfile
2021-11-12 14:42:27,787 INFO [conmon_server] Got a request: Request { metadata: MetadataMap { headers: {"te": "trailers", "content-type": "application/grpc", "user-agent": "tonic/0.6.1"} }, message: VersionRequest, extensions: Extensions }
Version: Response { metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 12 Nov 2021 19:42:27 GMT", "grpc-status": "0"} }, message: VersionResponse { version: "0.1.0" }, extensions: Extensions }
[pehunt@ovpn-114-123 conmon-rs(fork)]
 $ cat /proc/$(pidof conmon-server)/status | grep RSS
VmRSS:      2296 kB

```

note: a side effect of this PR is that conmon-server must be terminated with pkill or the like, because it is no longer a child of the calling terminal.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,11 +1,11 @@
 //! Configuration related structures
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, Result};
 use clap::{crate_version, Parser};
 use getset::{CopyGetters, Getters, Setters};
 use log::LevelFilter;
 use serde::{Deserialize, Serialize};
+use std::fs;
 use std::{env, path::PathBuf};
-use tokio::fs;
 
 macro_rules! prefix {
     () => {
@@ -74,20 +74,16 @@ impl Default for Config {
 
 impl Config {
     /// Validate the configuration integrity.
-    pub async fn validate(&self) -> Result<()> {
+    pub fn validate(&self) -> Result<()> {
         if !self.runtime().exists() {
             bail!("runtime path '{}' does not exist", self.runtime().display())
         }
 
         if self.socket().exists() {
-            fs::remove_file(self.socket())
-                .await
-                .context("remove existing socket file")?;
+            fs::remove_file(self.socket())?;
         }
         if let Some(parent) = self.socket().parent() {
-            fs::create_dir_all(parent)
-                .await
-                .context("ensure parent socket dir")?;
+            fs::create_dir_all(parent)?;
         }
 
         Ok(())
--- a/src/server.rs
+++ b/src/server.rs
@@ -10,7 +10,7 @@ use nix::{
     libc::_exit,
     unistd::{fork, ForkResult},
 };
-use std::{env, path::PathBuf};
+use std::{env, fs::File, io::Write, path::PathBuf};
 use tokio::{
     fs,
     net::UnixListener,
@@ -88,7 +88,11 @@ fn main() -> Result<(), Error> {
     // interrupt the child's execution.
     // 1: https://docs.rs/nix/0.23.0/nix/unistd/fn.fork.html#safety
     match unsafe { fork()? } {
-        ForkResult::Parent { child: _, .. } => {
+        ForkResult::Parent { child, .. } => {
+            if let Some(path) = server.config().conmon_pidfile() {
+                let child_str = format!("{}", child);
+                File::create(path)?.write_all(child_str.as_bytes())?;
+            }
             unsafe { _exit(0) };
         }
         ForkResult::Child => (),
--- a/src/server.rs
+++ b/src/server.rs
@@ -40,14 +40,10 @@ pub struct ConmonServerImpl {
 
 impl ConmonServerImpl {
     /// Create a new ConmonServerImpl instance.
-    pub async fn new() -> Result<Self> {
+    pub fn new() -> Result<Self> {
         let server = Self::default();
         server.init_logging().context("set log verbosity")?;
-        server
-            .config()
-            .validate()
-            .await
-            .context("validate config")?;
+        server.config().validate().context("validate config")?;
 
         server.init_self()?;
         Ok(server)
@@ -83,6 +79,9 @@ impl conmon::Server for ConmonServerImpl {
 }
 
 fn main() -> Result<(), Error> {
+    // First, initialize the server so we have access to the config pre-fork.
+    let server = ConmonServerImpl::new()?;
+
     // We need to fork as early as possible, especially before setting up tokio.
     // If we don't, the child will have a strange thread space and we're at risk of deadlocking.
     // We also have to treat the parent as the child (as described in [1]) to ensure we don't
@@ -92,18 +91,16 @@ fn main() -> Result<(), Error> {
         ForkResult::Parent { child: _, .. } => {
             unsafe { _exit(0) };
         }
-        ForkResult::Child => println!("I'm a new child process"),
+        ForkResult::Child => (),
     }
     // Use the single threaded runtime to save rss memory.
     let rt = runtime::Builder::new_current_thread().enable_io().build()?;
-    rt.block_on(start_server())?;
+    rt.block_on(start_server(server))?;
     Ok(())
 }
 
-async fn start_server() -> Result<(), Error> {
+async fn start_server(server: ConmonServerImpl) -> Result<(), Error> {
     let (shutdown_tx, shutdown_rx) = oneshot::channel();
-    let server = ConmonServerImpl::new().await?;
-
     let socket = server.config().socket().clone();
     tokio::spawn(start_sigterm_handler(socket, shutdown_tx));
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -6,6 +6,10 @@ use conmon_capnp::conmon;
 use futures::{AsyncReadExt, FutureExt};
 use getset::{Getters, MutGetters};
 use log::{debug, info};
+use nix::{
+    libc::_exit,
+    unistd::{fork, ForkResult},
+};
 use std::{env, path::PathBuf};
 use tokio::{
     fs,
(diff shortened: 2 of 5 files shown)


src/server.rs
```diff
@@ -83,12 +81,37 @@ impl Conmon for ConmonServerImpl {
     }
 }
 
-// Use the single threaded runtime to save rss memory
-#[tokio::main(flavor = "current_thread")]
-async fn main() -> Result<(), Error> {
-    let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
+fn main() {
```
cgwalters, line 84: You can keep the `-> Result<(), Error>` bits here though so that...

src/server.rs
```diff
@@ -83,12 +81,37 @@ impl Conmon for ConmonServerImpl {
...
+    // We also have to treat the parent as the child (as described in [1]) to ensure we don't
+    // interrupt the child's execution
+    // 1: https://docs.rs/nix/0.23.0/nix/unistd/fn.fork.html#safety
+    match unsafe{fork()} {
+        Ok(ForkResult::Parent { child, .. }) => {
+            if let Some(path) = server.config().conmon_pidfile() {
+                debug!("Writing child pid to pidfile {}", path.display());
+                let child_str = format!("{}", child);
+                File::create(path)
+                    .unwrap()
```
cgwalters, line 99: Then you can use `?` for a clean error instead of creating a core dump if we can't write the file.

Also similarly for the `fork()` invocation, it can just be `fork()?` and you don't need to match on `Ok/Err`.

src/server.rs
```diff
@@ -83,12 +81,37 @@ impl Conmon for ConmonServerImpl {
...
+fn main() {
+    // First, initialize the server so we have access to the config pre-fork.
+    let server = ConmonServerImpl::new().unwrap();
+
+    // We need to fork as early as possible, especially before setting up tokio.
+    // if we don't, the child will have a strange thread space and we're at risk of deadlocking.
+    // We also have to treat the parent as the child (as described in [1]) to ensure we don't
+    // interrupt the child's execution
+    // 1: https://docs.rs/nix/0.23.0/nix/unistd/fn.fork.html#safety
+    match unsafe{fork()} {
```
cgwalters, line 93: This looks like it could use a `cargo fmt`

cgwalters (commented):
Looks sane to me

src/server.rs
haircommander: unfortunately we still need the match for the fork case because we need to branch between parent/child of the fork, but it looks much cleaner for the other cases

src/server.rs
haircommander: It sure could :)

src/server.rs
cgwalters: Right, but you should be able drop the `Ok/Err` bits of that match by using `fork?`.

src/server.rs
```diff
@@ -83,12 +84,34 @@ impl Conmon for ConmonServerImpl {
...
+fn main() -> Result<(), Error> {
+    // First, initialize the server so we have access to the config pre-fork.
+    let server = ConmonServerImpl::new()?;
+
+    // We need to fork as early as possible, especially before setting up tokio.
+    // If we don't, the child will have a strange thread space and we're at risk of deadlocking.
+    // We also have to treat the parent as the child (as described in [1]) to ensure we don't
+    // interrupt the child's execution.
+    // 1: https://docs.rs/nix/0.23.0/nix/unistd/fn.fork.html#safety
+    match unsafe { fork() } {
```
rphillips, line 96: We should be able to call nix::unistd::daemon, right?

haircommander: I am starting to consider doing this, but right now we use the parent to create the pid file for the child which we couldn't do with daemon

haircommander:
this breaks our one integration test, I'm working on fixing it but may not before EOD

haircommander:
```
 $ ./target/release/conmon-server --runtime /usr/bin/runc --conmon-pidfile=/tmp/pidfile
2021-11-15 13:53:55,324 INFO [conmon_server] Set log level to INFO
2021-11-15 13:53:55,325 INFO [conmon_server::init] Missing sufficient privileges to adjust OOM score
[pehunt@toolbox conmon-rs(fork)]
 $ cat /proc/$(pidof conmon-server)/status | grep RSS
VmRSS:       196 kB
[pehunt@toolbox conmon-rs(fork)]
 $ ./target/release/conmon-client --runtime /usr/bin/runc --conmon-pidfile=/tmp/pidfile
2021-11-15 13:54:10,246 INFO [conmon_server] Got a request
received: 0.1.0
[pehunt@toolbox conmon-rs(fork)]
 $ cat /proc/$(pidof conmon-server)/status | grep RSS
VmRSS:       196 kB
```
lol this + capnproto makes conmon-rs miniscule

src/server.rs
```diff
@@ -78,12 +78,34 @@ impl conmon::Server for ConmonServerImpl {
...
```
saschagrunert, line 90:
```suggestion
    match unsafe { fork()? } {
```

Then check only for `ForkResult::Parent` and `ForkResult::Child` below.

tests/integration.rs
```diff
@@ -18,11 +21,23 @@ const CLIENT_BINARY: &str = "target/release/conmon-client";
...
     // Start the server
-    let mut server = Command::new(SERVER_BINARY).arg("--runtime=/tmp").spawn()?;
-    let pid = server.id().context("no pid for child")?;
+    let mut server = Command::new(SERVER_BINARY)
+        .arg("--runtime=/tmp")
+        .arg(pidfile_arg)
+        .spawn()?;
+
+    // Wait until parent has terminated.
+    server.wait().await.expect("wait for parent server process");
```
saschagrunert, line 34:
```suggestion
    server.wait().await?;
```

src/server.rs
```diff
@@ -78,12 +78,34 @@ impl conmon::Server for ConmonServerImpl {
...
+        Ok(ForkResult::Parent { child, .. }) => {
+            if let Some(path) = server.config().conmon_pidfile() {
+                let child_str = format!("{}", child);
+                File::create(path)?.write_all(child_str.as_bytes())?;
+            }
+            unsafe { _exit(0) };
```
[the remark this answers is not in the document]
haircommander: I can't figure out if it's [async signal safe](https://docs.rs/nix/0.23.0/nix/unistd/fn.fork.html#safety)

tests/integration.rs
```diff
@@ -18,11 +21,23 @@ const CLIENT_BINARY: &str = "target/release/conmon-client";
 
 #[tokio::test]
 async fn rss_verification() -> Result<()> {
+    let pidfile = NamedTempFile::new()?;
+    let mut pidfile_arg: String = "--conmon-pidfile=".to_owned();
+    pidfile_arg.push_str(pidfile.path().to_str().unwrap());
```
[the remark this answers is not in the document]
cgwalters: In the ostree crates we use https://docs.rs/camino/1.0.5/camino/ btw.

src/config.rs
```diff
@@ -74,20 +74,16 @@ impl Default for Config {
...
+    pub fn validate(&self) -> Result<()> {
         if !self.runtime().exists() {
             bail!("runtime path '{}' does not exist", self.runtime().display())
         }
 
         if self.socket().exists() {
-            fs::remove_file(self.socket())
-                .await
-                .context("remove existing socket file")?;
+            fs::remove_file(self.socket())?;
```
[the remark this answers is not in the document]
haircommander: because we don't want to start tokio until after the fork, but we want the information from the config before the fork. so we need to process the config fields synchronously

src/server.rs
haircommander: done!

tests/integration.rs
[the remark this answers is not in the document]
haircommander: unwrap in this case is to take the Option away, ? isn't working for me

src/server.rs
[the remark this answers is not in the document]
cgwalters: It's hard to know for sure; I don't think Rust makes strong promises about what system calls the standard library could make.  For example, `open()/write()/close()` corresponding to the `File::create().write_all()` calls above are safe...but in theory `std` could switch to some other system call.  Though that seems very unlikely.

As discussed on an internal chat I think really the best bet is to avoid a bare `fork()` at all and instead e.g. have podman use `systemd-run` or a direct call to the DBus `StartTransientUnit` API.  

Or to be really sure, only invoke `libc` API in the forked child - in the same way one would do in C.

But anyways, I think this code is all fine as is really too.  I'd keep the `_exit()` over `std::process::exit()`.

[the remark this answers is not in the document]
cgwalters: FWIW, the subprocess crate uses [fork and then raw exit](https://github.com/hniksic/rust-subprocess/blob/f092a654d6668736f0ce0f52fb31e17f3e31e091/src/popen.rs#L744).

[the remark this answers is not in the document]
haircommander: After more thought and discussion with @mrunalp , I think it'd be best to go forward with the daemonize code for now. it's been proven to work with conmon, and it will allow the port to happen in a cleaner way. in the future, we can revisit that decision

haircommander:
I believe this is ready @rphillips @mrunalp @saschagrunert

tests/integration.rs
```diff
@@ -18,11 +23,23 @@ const CLIENT_BINARY: &str = "target/release/conmon-client";
...
```
rphillips, line 28:
```diff
   diff --git a/tests/integration.rs b/tests/integration.rs
index de035d9d..49a4bb33 100644
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -24,8 +24,7 @@ const CLIENT_BINARY: &str = "target/release/conmon-client";
 #[tokio::test]
 async fn rss_verification() -> Result<()> {
     let pidfile = NamedTempFile::new()?;
-    let mut pidfile_arg: String = "--conmon-pidfile=".to_owned();
-    pidfile_arg.push_str(pidfile.path().to_str().unwrap());
+    let pidfile_arg = format!("--conmon-pidfile={}", pidfile.path().display());
     // Start the server
     let mut server = Command::new(SERVER_BINARY)
         .arg("--runtime=/tmp")

```

haircommander: one day I will understand rust string manipulation...

tests/integration.rs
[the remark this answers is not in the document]
rphillips: resolved in 33c937e

rphillips:
lgtm

@saschagrunert ?

src/config.rs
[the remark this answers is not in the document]
saschagrunert: Yeah that's fine. But the context adds additional information to the error message to better trace where the error is coming from. I think we can keep it for now, but later on we probably wanna add more `.context("…")` descriptions to our error messages to not trace "no such file or directory" message through the whole stack.

PR #37, opened 2021-11-16
Add go client and library

haircommander:
relies on https://github.com/containers/conmon-rs/pull/28

This PR adds a go client for communicating with the conmon-server. Since all existing container managers are written in go, it's not as useful to test and work with a rust client. Rather, creating a go client communicating through UDS is a more realistic approach.

The idea is that we create a library that acts as the go client. eventually, as more methods are added, we test and extend the go client. Eventually, when we're ready to use conmon-rs, we have most of the go work done, and container managers can use the built in library as expected.


haircommander:
builds on https://github.com/containers/conmon-rs/pull/38 now, but it's ready whenever that merges @rphillips @saschagrunert

pkg/client/client.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	if config.Stdout != nil {
+		cmd.Stdout = os.Stdout
+	}
+	if config.Stderr != nil {
+		cmd.Stderr = os.Stderr
+	}
+	if err := cmd.Run(); err != nil {
+		return nil, err
+	}
```
rphillips, line 56: iirc we need a cmd.Wait() deferred

pkg/client/client_test.go
```diff
@@ -0,0 +1,113 @@
...
+		pidFile, err := ioutil.TempFile("", "pidfile")
+		Expect(err).To(BeNil())
+		pidFilePath = pidFile.Name()
+		socket, err := ioutil.TempFile("", "socket")
+		Expect(err).To(BeNil())
+		socketPath = socket.Name()
+	})
+
+	AfterEach(func() {
+		Expect(os.Remove(socketPath)).To(BeNil())
```
rphillips, line 40: should remove the pidFile as well

haircommander: pidfile is removed by the client itself. We could verify it's removed, or drop that behavior? I am unsure of it tbh

pkg/client/client.go
haircommander: run is Start() + Wait() :)

rphillips: ah hah :)

saschagrunert (approved):

PR #42, opened 2021-11-19
Add unit tests to init module

saschagrunert:
To be able to test the init functions, we require a wrapping structure
and replace-able implementation for it. This has been now added, while
the default init implementation is being used from the server.

### DIFF
--- a/conmon-rs/server/src/init.rs
+++ b/conmon-rs/server/src/init.rs
@@ -1,28 +1,159 @@
 use anyhow::{bail, Result};
-use libc::{setlocale, LC_ALL};
+use libc::{c_char, c_int, setlocale, LC_ALL};
 use log::info;
-use std::ffi::CString;
-use std::fs::File;
-use std::io::{ErrorKind, Write};
+use std::{
+    ffi::CString,
+    fs::File,
+    io::{self, ErrorKind, Write},
+    path::Path,
+};
 
-/// Unset the locale for the current process.
-pub fn unset_locale() -> Result<()> {
-    unsafe {
-        setlocale(LC_ALL, CString::new("")?.as_ptr());
-    }
-    Ok(())
+#[cfg(test)]
+use mockall::{automock, predicate::*};
+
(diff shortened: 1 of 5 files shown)


conmon-rs/server/src/init.rs
```diff
@@ -1,24 +1,172 @@
...
+    #[derive(Default)]
+    struct TestInit {
+        set_locale_called: Option<LocaleCategory>,
+        create_file_called: Option<PathBuf>,
+        create_file_fails: bool,
+        write_all_called: Option<Vec<u8>>,
+        write_all_fails: Option<ErrorKind>,
+    }
+
+    impl InitImpl for TestInit {
```
rphillips, line 90: mockall crate seems like a good mock library... it would remove the boilerplate here.

saschagrunert: Thanks, changed as suggested. PTAL again

rphillips (approved):

conmon-rs/server/src/init.rs
rphillips: wow... wasn't expecting it to be this clean, but it looks great.

conmon-rs/server/src/init.rs
```diff
@@ -1,24 +1,161 @@
...
+where
+    T: InitImpl,
+{
+    /// Unset the locale for the current process.
+    pub fn unset_locale(&self) {
+        self.imp.setlocale(LocaleCategory::LcAll, "");
+    }
+
+    /// Helper to adjust the OOM score of the currently running process.
+    pub fn set_oom<S: AsRef<str>>(&self, score: S) -> Result<()> {
```
mrunalp, line 28: nit: rename to `set_oom_score`

saschagrunert: Renamed :+1:

haircommander:
we seem to need another rebase

saschagrunert:
Rebased again.

haircommander:
LGTM

PR #44, opened 2021-11-19
enables multi threaded scheduler again

rphillips:
With the migration to capnproto, I enabled the multithreaded tokio scheduler again. RSS is at 1.4 MB on my machine. Might be worth it based on todays discussion on managing multiple log streams.


rphillips:
discussed in slack

PR #53, opened 2021-11-19
remove dependency on gettext-rs

### LINKED ISSUE #43: gettext-sys compile times
Is there a way to get rid of gettext-sys? We use it for setting the locale, but the compile times with it seem to be quite long.

rphillips:
Fixes #43

### DIFF
--- a/conmon-rs/server/src/init.rs
+++ b/conmon-rs/server/src/init.rs
@@ -1,12 +1,16 @@
 use anyhow::{bail, Result};
-use gettextrs::{setlocale, LocaleCategory};
+use libc::{setlocale, LC_ALL};
 use log::info;
+use std::ffi::CString;
 use std::fs::File;
 use std::io::{ErrorKind, Write};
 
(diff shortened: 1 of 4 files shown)


conmon-rs/server/src/init.rs
```diff
@@ -1,12 +1,16 @@
...
 use std::io::{ErrorKind, Write};
 
 /// Unset the locale for the current process.
-pub fn unset_locale() {
-    setlocale(LocaleCategory::LcAll, "");
+pub fn unset_locale() -> Result<()> {
+    unsafe {
+        setlocale(LC_ALL, CString::new("")?.as_ptr());
+    }
+    Ok(())
```
mrunalp, line 13: Do we need to return this since we aren't actually propagating any errors?

rphillips: It does get propagated in this case.

haircommander:
LGTM

PR #54, opened 2021-11-19
Add CreateContainer RPC

haircommander:
as well as add a test that calls it
TODO:
- [x] actually check there's a container in the runtime
- [ ] propegate status in a sane way (probably in a follow up)
- [x] make the tests work with runc:
    - we get an odd error `time="2021-11-19T17:11:43-05:00" level=error msg="runc create failed: unable to start container process: error during container init: unable to store init state: %!w(<nil>)"`


pkg/client/client_test.go
```diff
@@ -25,55 +42,103 @@ func TestConmonClient(t *testing.T) {
...
-		Expect(err).To(BeNil())
-		socketPath = socket.Name()
+		tmpDir = MustTempDir("conmon-client")
+		pidFilePath = MustFileInTempDir(tmpDir, "pidfile")
+		socketPath = MustFileInTempDir(tmpDir, "socket")
 	})
 
 	AfterEach(func() {
-		Expect(os.Remove(socketPath)).To(BeNil())
+		Expect(os.RemoveAll(tmpDir)).To(BeNil())
```
haircommander, line 54: hm this makes the tests fail...

conmon-rs/server/src/rpc.rs
```diff
@@ -16,4 +18,26 @@ impl conmon::Server for Server {
...
+            "Got a create container request for id {}",
+            pry!(req.get_id())
+        );
+        let args = self.generate_runtime_args(&params).unwrap();
+        let mut child = Command::new(self.config().runtime()).args(args).spawn().unwrap();
+        let pid = child.id().unwrap();
+        let id = req.get_id().unwrap().to_string();
+        tokio::spawn(async move {
+            let status = child.wait().await;
+            println!("status for container ID {} is {}", id, status.unwrap());
```
rphillips, line 38: debug! probably

conmon-rs/server/src/rpc.rs
```diff
@@ -16,4 +18,26 @@ impl conmon::Server for Server {
...
         Promise::ok(())
     }
+
+    fn create_container(
+        &mut self,
+        params: conmon::CreateContainerParams,
+        mut results: conmon::CreateContainerResults,
+    ) -> Promise<(), capnp::Error> {
+        let req = pry!(pry!(params.get()).get_request());
+        info!(
```
rphillips, line 28: debug!

conmon-rs/server/src/rpc.rs
```diff
@@ -16,4 +18,26 @@ impl conmon::Server for Server {
...
+            "Got a create container request for id {}",
+            pry!(req.get_id())
+        );
+        let args = self.generate_runtime_args(&params).unwrap();
+        let mut child = Command::new(self.config().runtime()).args(args).spawn().unwrap();
```
rphillips, line 33: I think we will want/need the proper error handling here without the unwrap.

rphillips: Generally, we should avoid unwrap() since they can cause the process to crash.

mrunalp:
`time="2021-11-19T17:11:43-05:00" level=error msg="runc create failed: unable to start container process: error during container init: unable to store init state: %!w(<nil>)"`

That looks a permission issue writing it. Are you able to successfully manually `runc create` with the rootfs + config.json?

conmon-rs/server/src/rpc.rs
```diff
@@ -16,4 +18,26 @@ impl conmon::Server for Server {
...
+        let pid = child.id().unwrap();
+        let id = req.get_id().unwrap().to_string();
+        tokio::spawn(async move {
```
mrunalp, line 36: We will eventually want to send this pid over to a loop monitoring pid fds. (Not blocking for now.)

pkg/client/client.go
```diff
@@ -119,6 +119,45 @@ func (c *ConmonClient) Version(ctx context.Context) (string, error) {
...
+	ID         string
+	BundlePath string
+}
+
+func (c *ConmonClient) CreateContainer(ctx context.Context, cfg *CreateContainerConfig) (uint32, error) {
+	conn, err := c.newRPCConn()
+	if err != nil {
+		return 0, err
+	}
+	client := proto.Conmon{Client: conn.Bootstrap(context.Background())}
```
rphillips, line 132: this should be `ctx`

conmon-rs/server/src/rpc.rs
```diff
@@ -12,8 +15,50 @@ impl conmon::Server for Server {
...
+    ) -> Promise<(), capnp::Error> {
+        let req = pry!(pry!(params.get()).get_request());
+        debug!(
+            "Got a create container request for id {}",
+            pry!(req.get_id())
+        );
+        let args = match self.generate_runtime_args(&params) {
+            Ok(x) => x,
+            Err(e) => return Promise::err(e),
+        };
```
saschagrunert, line 36:
```suggestion
        let args = pry!(self.generate_runtime_args(&params));
```

conmon-rs/server/src/rpc.rs
```diff
@@ -12,8 +15,50 @@ impl conmon::Server for Server {
...
+
+        let mut child = match Command::new(self.config().runtime()).args(args).spawn() {
+            Ok(x) => x,
+            Err(e) => return Promise::err(e.into()),
+        };
```
saschagrunert, line 41:
```suggestion
        let mut child = pry!(Command::new(self.config().runtime()).args(args).spawn());
```

conmon-rs/server/src/rpc.rs
```diff
@@ -12,8 +15,50 @@ impl conmon::Server for Server {
...
+            None => {
+                return Promise::err(Error::failed(
+                    "Child of container creation had none id".to_string(),
+                ))
+            }
+        };
+        let id = match req.get_id() {
+            Ok(x) => x.to_string(),
+            Err(e) => return Promise::err(e),
+        };
```
saschagrunert, line 54:
```suggestion       
        let id = pry!(req.get_id()).to_string();
```

conmon-rs/server/src/rpc.rs
```diff
@@ -12,8 +15,50 @@ impl conmon::Server for Server {
...
+
+        let pid = match child.id() {
+            Some(x) => x,
+            None => {
+                return Promise::err(Error::failed(
+                    "Child of container creation had none id".to_string(),
+                ))
+            }
+        };
```
saschagrunert, line 50:
```suggestion
        let pid = pry!(child
            .id()
            .ok_or_else(|| Error::failed("Child of container creation had none id".into())));
```

conmon-rs/server/src/rpc.rs
haircommander: goodness I tried for so long to get something like that to work. thank you :upside_down_face:

haircommander:
:heavy_check_mark: tests pass PTAL @saschagrunert @rphillips

haircommander:
I've updated runtimeRoot handling to be a cli flag instead of passed as proto. I figure, similarly to the runtime path, that a single pod will have the same runtime root for all creation and exec operations. [PreCommandOpts and PostCommandOpts](https://github.com/haircommander/conmon-rs/commit/588fa43c5bf92f985a1a410b634d29b0b41f9b88) can be added later

conmon-rs/server/src/config.rs
```diff
@@ -79,6 +89,14 @@ impl Config {
             bail!("runtime path '{}' does not exist", self.runtime().display())
         }
 
+        if let Some(rr) = self.runtime_root() {
+            if rr.exists() {
+                fs::create_dir_all(rr)?;
```
saschagrunert, line 94: Why do we have to create the dir if it already exist?

saschagrunert (approved):
Just a nit, otherwise LGTM

conmon-rs/server/src/config.rs
haircommander: because I write buggy code :sweat_smile: good catch

saschagrunert (approved):

PR #57, opened 2021-11-22
Implement k8s file logger

saschagrunert:
This adds the kubernetes log file functionality as dedicated module.

### DIFF
--- a/conmon-rs/server/src/cri_logger.rs
+++ b/conmon-rs/server/src/cri_logger.rs
@@ -1,0 +1,256 @@
+//! File logging functionalities.
+
+#![allow(dead_code)] // TODO: remove me when actually used
+
+use anyhow::{Context, Result};
+use chrono::offset::Local;
+use getset::{CopyGetters, Getters, MutGetters, Setters};
+use log::{debug, trace};
+use memchr::memchr;
+use std::{
+    marker::Unpin,
+    path::{Path, PathBuf},
+};
+use tokio::{
+    fs::{File, OpenOptions},
+    io::{AsyncBufRead, AsyncBufReadExt, AsyncWriteExt, BufReader, BufWriter},
+};
+
+#[derive(Debug, CopyGetters, Getters, MutGetters, Setters)]
+/// The main structure used for container log handling.
+pub struct CriLogger {
+    #[getset(get)]
+    /// Path to the file on disk.
+    path: PathBuf,
+
+    #[getset(get, get_mut, set)]
+    /// Open file handle of the `path`.
+    file: BufWriter<File>,
+
+    #[getset(get_copy)]
+    /// Maximum allowed log size in bytes.
+    max_log_size: Option<usize>,
+}
+
+#[derive(Clone, Copy, Debug)]
+/// Available pipe types.
+pub enum Pipe {
+    /// Standard output.
+    StdOut,
+
+    /// Standard error.
+    StdErr,
+}
+
+impl CriLogger {
+    /// Create a new file logger instance.
+    pub async fn from<T: AsRef<Path>>(path: T, max_log_size: Option<usize>) -> Result<Self> {
+        Ok(Self {
+            path: path.as_ref().into(),
+            file: Self::open(&path).await?,
+            max_log_size,
+        })
+    }
+
+    /// Write the contents of the provided reader into the file logger. Ensure to manually call
+    /// `flush` if you need the written log data on disk.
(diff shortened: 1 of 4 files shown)


conmon-rs/server/src/file_logger.rs
```diff
@@ -0,0 +1,256 @@
+//! File logging functionalities.
```
haircommander, line 1: we should probably make clear this is the cri format, as json-file may one day be a supported type

saschagrunert: I changed the module to be `cri_logger.rs` does this fit better?

conmon-rs/server/src/cri_logger.rs
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Reopen the container log file.
+    pub async fn reopen(&mut self) -> Result<()> {
+        debug!("Reopen container log {}", self.path().display());
+        self.file().get_ref().sync_all().await?;
+        self.set_file(Self::open(self.path()).await?);
```
rphillips, line 131: I think we need to drop self.file first, so the Self::open doesn't open the already opened file.

conmon-rs/server/src/cri_logger.rs
```diff
@@ -0,0 +1,256 @@
...
+        let mut sut = CriLogger::from(path, None).await?;
+
+        sut.write(Pipe::StdOut, &mut reader).await?;
+        sut.flush().await?;
+
+        let res = fs::read_to_string(path)?;
+        assert!(res.contains(" stdout F this is a line"));
+        assert!(res.contains(" stdout F and another line"));
+
+        DateTime::parse_from_rfc3339(res.split_whitespace().nth(0).context("no timestamp")?)
```
rphillips, line 200: linter says nth(0) should be next()

conmon-rs/server/src/cri_logger.rs
saschagrunert: We cannot drop self.file because then we would partially move content out of `self`, which is only borrowed as mutable reference.

```diff
diff --git a/conmon-rs/server/src/cri_logger.rs b/conmon-rs/server/src/cri_logger.rs
index e34da9d7..b891b4fe 100644
--- a/conmon-rs/server/src/cri_logger.rs
+++ b/conmon-rs/server/src/cri_logger.rs
@@ -128,6 +128,7 @@ impl CriLogger {
     pub async fn reopen(&mut self) -> Result<()> {
         debug!("Reopen container log {}", self.path().display());
         self.file().get_ref().sync_all().await?;
+        drop(self.file);
         self.set_file(Self::open(self.path()).await?);
         Ok(())
     }
```

```
error[E0507]: cannot move out of `self.file` which is behind a mutable reference
   --> conmon-rs/server/src/cri_logger.rs:131:14
    |
131 |         drop(self.file);
    |              ^^^^^^^^^ move occurs because `self.file` has type `tokio::io::BufWriter<tokio::fs::File>`, which does not implement the `Copy` trait
```

`self.file` gets automatically dropped on replacement because it runs out of scope. For example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7d454dac7b772a4e5c05e98a2f383e21

conmon-rs/server/src/cri_logger.rs
saschagrunert: Fixed, thank you!

conmon-rs/server/src/cri_logger.rs
rphillips: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5d77116cd57ac89e845ddb1844b2a368

This example opens the file, and then drops the reference.

rphillips: Perrhaps https://docs.rs/io-close/latest/io_close/

cgwalters: I think for this scenario you need to make it an `Option<File>`.  Then it's just:
```
self.file = None;
self.set_file(...)
```

rphillips: Perhaps my concern here isn't warranted. The kubelet does the rotation and removes the old file, so the open is opening a completely new file.

cgwalters: I'd agree with that.

rphillips: resolving... since the kubelet removes the file and then requests for a new one to be opened.

rphillips (approved):

PR #65, opened 2021-11-23
Add console socket

saschagrunert:
A few TODOs are still open but the basic fd passing works for now.

### DIFF
--- a/conmon-rs/server/src/console.rs
+++ b/conmon-rs/server/src/console.rs
@@ -1,0 +1,207 @@
+//! Console socket functionalities.
+
+#![allow(dead_code)] // TODO: remove me when actually used
+
+use anyhow::{bail, format_err, Context, Result};
+use getset::Getters;
+use log::{debug, error, trace};
+use nix::sys::termios::{self, OutputFlags, SetArg};
+use sendfd::RecvWithFd;
+use std::{
+    io::ErrorKind,
+    os::unix::{fs::PermissionsExt, io::RawFd},
+    path::{Path, PathBuf},
+    sync::{Arc, RwLock},
+    thread::{self, JoinHandle},
+};
+use tempfile::Builder;
+use tokio::{
+    fs,
+    io::{AsyncWriteExt, Interest},
+    net::{UnixListener, UnixStream},
+    runtime::Runtime,
+};
+
+#[derive(Debug, Getters)]
+#[getset(get)]
+pub struct Console {
+    path: PathBuf,
+    handle: JoinHandle<Result<()>>,
+    fd: Arc<RwLock<Option<RawFd>>>,
+}
+
+impl Console {
+    /// Setup a new console socket.
+    pub fn new() -> Result<Self> {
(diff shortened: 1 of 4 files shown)


conmon-rs/server/src/console.rs
```diff
@@ -0,0 +1,119 @@
...
+                        }
+                        Err(e) => {
+                            return Err(e.into());
+                        }
+                    }
+                }
+            }
+        };
+
+        let (left, right) = tokio::join!(sut.start(), verify);
```
saschagrunert, line 113: @haircommander is this usable? I mean the whole async can run in parallel, but we would have to join the tasks together and the one from the `Console` is blocking on `await`.

haircommander: the goal is to be able to have the container i/o processed in parallel with other i/o processing and container creation. as long as a user's terminal remains responsive, and conmon-rs can make foward progress simultaneously, I think we're good. I am having trouble deciding if this code block does that by looking at it though

saschagrunert: I'll do some more experiments and unmark the WIP when ready for review.

haircommander (changes requested):
Now that we actually create a container, wdyt about wiring this up? If you'd prefer as a follow up that's fine with me

saschagrunert:
> Now that we actually create a container, wdyt about wiring this up? If you'd prefer as a follow up that's fine with me

Yeah I can follow up on that as well as those TODOs :)

conmon-rs/server/src/console.rs
```diff
@@ -0,0 +1,205 @@
...
+        })
+    }
+
+    async fn handle_fd_receive(
+        mut stream: UnixStream,
+        path: &Path,
+        fd_state: Arc<RwLock<Option<RawFd>>>,
+    ) -> Result<()> {
+        loop {
+            if stream.ready(Interest::READABLE).await?.is_readable() {
```
haircommander, line 94: what if we made this `if !stream.ready(Interest::READABLE).await?.is_readable() { continue; }` and then we could  move the rest of the block in?

saschagrunert: Sound good, changed as suggested :+1:

conmon-rs/server/src/console.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    async fn handle_fd_receive(
+        mut stream: UnixStream,
+        path: &Path,
+        fd_state: Arc<RwLock<Option<RawFd>>>,
+    ) -> Result<()> {
+        loop {
+            if !stream.ready(Interest::READABLE).await?.is_readable() {
+                continue;
+            }
```
haircommander, line 96: something to think about is whether we can optimize this a bit. obviously we don't want to miss any data from the fd or cause the user to stall, but it seems like we're busy looping here and that may cause unneeded cpu usage

haircommander:
LGTM, needs rebase though..

saschagrunert:
> LGTM, needs rebase though..

Rebased.

rphillips (approved):

PR #75, opened 2021-12-01
Add terminal support on container creation

saschagrunert:
We now wire-in the console module for container creation.


pkg/client/client.go
```diff
@@ -143,6 +143,7 @@ func (c *ConmonClient) CreateContainer(ctx context.Context, cfg *CreateContainer
 		if err := req.SetBundlePath(cfg.BundlePath); err != nil {
 			return err
 		}
+		req.SetTerminal(true)
```
haircommander, line 146: can we add a CreateContainerConfig field for this?

pkg/client/client_test.go
```diff
@@ -239,6 +239,7 @@ func generateRuntimeConfig(bundlePath, rootfs string) error {
 		return err
 	}
 	g.SetProcessCwd("/")
+	g.SetProcessTerminal(true)
```
haircommander, line 242: it would be cool if we could test it fails if SetProcessTerminal(false) but SetTerminal(true) and vice versa. if it doesn't fail we don't have to look that hard into it for now

pkg/client/client.go
saschagrunert: Yes

pkg/client/client_test.go
saschagrunert: Added a new test case for the terminal, so now we have two :+1:

haircommander:
LGTM

conmon-rs/server/src/console.rs
```diff
@@ -23,32 +25,47 @@ use tokio::{
...
-    /// Returns true if the console socket is successfully connected.
-    pub fn is_connected(&self) -> bool {
-        self.fd().read().map(|fd| fd.is_some()).unwrap_or_else(|e| {
-            error!("Unable to retrieve fd lock: {}", e);
-            false
+        let (ready_tx, ready_rx) = mpsc::channel();
+        let (connected_tx, connected_rx) = mpsc::channel();
+
+        thread::spawn(move || Self::listen(&path_clone, fd_clone, ready_tx, connected_tx));
+        ready_rx.recv().context("wait for ready to be listener")?;
```
rphillips, line 51:
```suggestion
        ready_rx.recv().context("wait for listener to be ready")?;
```

rphillips (changes requested):

rphillips (approved):

PR #76, opened 2021-12-01
server: return correct pid file

haircommander:
before we were returning the runtime process pid file, not the container's


conmon-rs/server/src/rpc.rs
```diff
@@ -37,33 +38,44 @@ impl conmon::Server for Server {
...
         results.get().init_response().set_container_pid(pid);
         Promise::ok(())
     }
 }
+
+fn pidfile_from_params(params: &conmon::CreateContainerParams) -> capnp::Result<String> {
+    let mut pidfile_pathbuf = PathBuf::from(params.get()?.get_request()?.get_bundle_path()?);
+    pidfile_pathbuf.push("pidfile");
+
+    let pidfile_str = pidfile_pathbuf.into_os_string();
```
rphillips, line 74: nit: this line and the next line can be chained all the way through.

conmon-rs/server/src/rpc.rs
```diff
@@ -37,33 +38,44 @@ impl conmon::Server for Server {
...
+        let pid = pry!(pry!(fs::read_to_string(pidfile))
+            .parse::<u32>()
+            .map_err(|e| Error::failed(format!("{:#}", e))));
+
         results.get().init_response().set_container_pid(pid);
         Promise::ok(())
     }
 }
+
+fn pidfile_from_params(params: &conmon::CreateContainerParams) -> capnp::Result<String> {
```
saschagrunert, line 70: If we would return a `PathBuf` here, then we could simplify the whole logic abit. For example:

```diff
diff --git a/conmon-rs/server/src/lib.rs b/conmon-rs/server/src/lib.rs
index f7e90b41..f52ea4ff 100644
--- a/conmon-rs/server/src/lib.rs
+++ b/conmon-rs/server/src/lib.rs
@@ -156,11 +156,11 @@ impl Server {
     }
 
     /// Generate the OCI runtime CLI arguments from the provided parameters.
-    fn generate_runtime_args(
+    fn generate_runtime_args<T: AsRef<Path>>(
         &self,
         params: &conmon::CreateContainerParams,
         maybe_console: &Option<Console>,
-        pidfile: &String,
+        pidfile: T,
     ) -> Result<Vec<String>> {
         let req = params.get()?.get_request()?;
         let id = req.get_id()?.to_string();
@@ -176,7 +176,7 @@ impl Server {
             "--bundle".to_string(),
             bundle_path,
             "--pid-file".to_string(),
-            pidfile.to_string(),
+            pidfile.as_ref().display().to_string(),
         ]);
 
         if let Some(console) = maybe_console {
diff --git a/conmon-rs/server/src/rpc.rs b/conmon-rs/server/src/rpc.rs
index 46a53da8..be68cc70 100644
--- a/conmon-rs/server/src/rpc.rs
+++ b/conmon-rs/server/src/rpc.rs
@@ -67,15 +67,10 @@ impl conmon::Server for Server {
     }
 }
 
-fn pidfile_from_params(params: &conmon::CreateContainerParams) -> capnp::Result<String> {
+fn pidfile_from_params(params: &conmon::CreateContainerParams) -> capnp::Result<PathBuf> {
     let mut pidfile_pathbuf = PathBuf::from(params.get()?.get_request()?.get_bundle_path()?);
     pidfile_pathbuf.push("pidfile");
 
-    let pidfile_str = pidfile_pathbuf.into_os_string();
-    let pidfile_str: String = pidfile_str
-        .to_str()
-        .ok_or_else(|| capnp::Error::failed(format!("Failed to convert pidfile to string")))?
-        .into();
-    debug!("pidfile is {}", pidfile_str);
-    Ok(pidfile_str)
+    debug!("pidfile is {}", pidfile_pathbuf.display());
+    Ok(pidfile_pathbuf)
 }
```

haircommander: that is simpler! fixed

conmon-rs/server/src/rpc.rs
haircommander: dropped as part of fix below

saschagrunert (approved):

PR #77, opened 2021-12-02
Use helper macro for mapping errors

saschagrunert:
We will have a need for mapping the anyhow::Error to the capnp
promise-based error more than often. For we now introduce a new
macro for convenience reasons.


saschagrunert:
We may find a better binary because we often hit:
```
<*url.Error | 0xc000212090>: {
            Op: "Get",
            URL: "https://busybox.net/downloads/binaries/1.31.0-i686-uclibc/busybox",
            Err: <http.tlsHandshakeTimeoutError>{},
        }
```

haircommander:
great idea!

> We may find a better binary location because we often hit:
yeah that's a good point. or we could cache the binary in the ci setup code and retry on handshake failure..

PR #78, opened 2021-12-02
Add reading from console pty

saschagrunert:
Playing around with a Pty abstraction to be able to read/write to the master pty.

### DIFF
--- a/conmon-rs/server/src/console.rs
+++ b/conmon-rs/server/src/console.rs
@@ -1,5 +1,6 @@
 //! Console socket functionalities.
 
+use crate::iostreams::IOStreams;
 use anyhow::{bail, format_err, Context, Result};
 use getset::Getters;
 use log::{debug, error, trace};
@@ -9,10 +10,8 @@ use std::{
     io::ErrorKind,
     os::unix::{fs::PermissionsExt, io::RawFd},
     path::{Path, PathBuf},
-    sync::{
-        mpsc::{self, Receiver, Sender},
-        Arc, RwLock,
-    },
+    str,
+    sync::mpsc::{self, Receiver, Sender},
     thread,
     time::Duration,
 };
@@ -30,9 +29,6 @@ pub struct Console {
     path: PathBuf,
 
     connected_rx: Receiver<()>,
-
-    #[allow(dead_code)]
-    fd: Arc<RwLock<Option<RawFd>>>,
 }
 
 impl Console {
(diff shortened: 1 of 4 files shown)


conmon-rs/server/src/console.rs
```diff
@@ -177,6 +160,18 @@ impl Console {
...
+        debug!("Start reading from pty");
+        let mut reader = BufReader::new(pty);
+        loop {
+            let mut buf = vec![];
+            match reader.read(&mut buf) {
+                Ok(n) => debug!("Read {} bytes: {:?}", n, buf),
+                Err(e) => error!("Unable to read from terminal: {}", e),
+            }
+        }
+    }
```
saschagrunert, line 174: I'm not sure about this, like how should we read the pty? Line-wise?

saschagrunert:
@haircommander PTAL

conmon-rs/server/src/pty.rs
```diff
@@ -0,0 +1,222 @@
...
+        let mode = Mode::from_bits_truncate(0o777);
+        const DEV_NULL: &str = "/dev/null";
+
+        if !self.worker_stdin().is_set() {
+            let fd = OpenOptions::new().read(true).open(DEV_NULL)?.into_raw_fd();
+            unistd::dup2(fd, libc::STDIN_FILENO).context("dup over stdin")?;
+            self.set_worker_stdin(fd.into());
+        } else {
+            stat::fchmod(libc::STDIN_FILENO, mode).context("chmod stdin")?
+        }
```
haircommander, line 132: is a helper function here possible?

conmon-rs/server/src/pty.rs
```diff
@@ -0,0 +1,222 @@
...
+#[derive(Clone, Copy, Debug, CopyGetters, Setters)]
+#[getset(get_copy, set)]
+pub struct Pty {
+    main_stdin: Stream,
+    main_stderr: Stream,
+    main_stdout: Stream,
+    worker_stdin: Stream,
+    worker_stderr: Stream,
+    worker_stdout: Stream,
+}
```
haircommander, line 27: do you have an idea how !terminal stream handling will work? if it will use this struct (as I see we're using these fields when dup'ing), maybe it should have a different name, as it won't actually be a pty

saschagrunert: My idea was to make it technically work in the same way as conmon is doing it today. Do you think there is an easier or better way to do it? (if so, let's sync on this topic)

I was thinking about a better name for `Pty`. Using `IOStreams` for now, maybe you have something better in mind.

conmon-rs/server/src/pty.rs
saschagrunert: Yeah, it would make things harder to read since there are slight differences between the arguments used in the functions. I decided to leave it as-is for now.

conmon-rs/server/src/pty.rs
haircommander: yeah IOStreams is good, and I +1 the idea of keeping in sync with old conmon

saschagrunert:
Rebased to make the CI happy.

haircommander:
we are currently printing tons of `ERROR [conmon_server::iostreams] Unable to read from terminal: stream not set` in the integration tests, I don't know if we should log there

saschagrunert:
I have to revisit this PR in the next couple of days.

saschagrunert:
@haircommander PTAL

haircommander:
LGTM

PR #80, opened 2021-12-02
monitor children and write exit paths

haircommander:
This set of patches encompasses child handling in conmon-rs. specifically it:
- adds state to keep track of the grandchildren container processes
- sets self as subreaper to catch grandchildren exiting
- catches the sigchld from the children/grandchildren and writes their exit codes down
- kills the children its tracking upon exiting

you know, like a *normal* parent does :slightly_smiling_face:


conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl ChildReaper {
+    pub fn init_reaper(&self) -> Result<()> {
+        let mut mask = SigSet::empty();
+        mask.add(SIGCHLD);
+        mask.thread_block()?;
+        // Signals are queued up on the file descriptor
+        let mut sfd = SignalFd::with_flags(&mask, SfdFlags::empty())?;
+
+        let children = self.children.clone();
```
cgwalters, line 23: This is a stylistic thing, but I prefer doing:
```suggestion
        let children = Arc::clone(self.children);
```

The rationale here is that `.clone()` can do almost anything - for example, you could be copying all of the backing storage behind a gigantic hash map.  In this case we're just bumping a reference count.  But to figure that out, I had to go look.  By using `Arc::clone()` we make that very explicit to the reader that this isn't actually an *expensive* clone.

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,124 @@
...
+        thread::spawn(move || {
+            let children = children.clone();
```
cgwalters, line 25: Is this extra reference really needed?

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,124 @@
...
+};
+use nix::unistd::Pid;
+use std::path::PathBuf;
+use std::thread;
+use std::{collections::HashMap, fs::File, io::Write, sync::Arc, sync::Mutex};
+
+impl ChildReaper {
+    pub fn init_reaper(&self) -> Result<()> {
+        let mut mask = SigSet::empty();
+        mask.add(SIGCHLD);
```
cgwalters, line 18: BTW, see https://github.com/containers/containers-image-proxy-rs/pull/13 in which I discovered that tokio wants to install its own `SIGCHLD` handler.  And I think using `signalfd()` here would conflict with that.  

This is all probably fine but be aware that things will go wrong if you end up spawning a process [using tokio](https://docs.rs/tokio/1.14.0/tokio/process/struct.Command.html).

(The long term fix is to use pidfd - which, actually why aren't you using pidfd here?)

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,124 @@
...
+        let children = self.children.clone();
+        thread::spawn(move || {
+            let children = children.clone();
+            loop {
+                let children = children.clone();
+                match sfd.read_signal() {
+                    // we caught a signal
+                    Ok(Some(sig)) => {
+                        // Immediately spawn a thread to reduce risk of dropping
+                        // a SIGCHLD.
```
cgwalters, line 32: Ah...wow yeah, I think using pidfd would be cleaner.

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        let reapable_child = ReapableChild {
+            id: id.clone(),
+            child: c.into(),
+            exit_paths: exit_paths,
+        };
+        if let Some(old) = map.insert(c, reapable_child) {
+            bail!("Repeat PID for container {} found", old.id);
+        }
+        drop(map);
```
cgwalters, line 90: This doesn't seem necessary.

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+        Ok(())
+    }
+
+    pub fn add_child(&self, id: String, c: i32, exit_paths: Vec<PathBuf>) -> Result<()> {
+        let locked_children = self.children.clone();
+        let mut map = locked_children.lock().unwrap();
+
+        let reapable_child = ReapableChild {
+            id: id.clone(),
```
cgwalters, line 83: We're taking ownership of `id`, so this `clone()` isn't necessary.

conmon-rs/server/src/child_reaper.rs
cgwalters: e.g. https://github.com/containers/containers-image-proxy-rs/blob/408dbded66f1bd2556cb697bb82f842eb4e401ae/src/imageproxy.rs#L281

conmon-rs/server/src/lib.rs
```diff
@@ -106,27 +120,38 @@ impl Server {
...
+                handled_sig = Signal::SIGINT;
             }
         };
 
         let _ = shutdown_tx.send(());
+
+        // TODO FIXME Ideally we would drop after socket file is removed,
+        // but the removal is taking longer than 10 seconds, indicating someone
+        // is keeping it open...
+        reaper.kill_children(handled_sig)?;
```
cgwalters, line 148: BTW do we need this versus `PRCTL_PR_SET_PDEATHSIG`?  See e.g. https://github.com/containers/containers-image-proxy-rs/pull/16

conmon-rs/server/src/child_reaper.rs
haircommander: does rhel 8 support pidfds? I didn't implement with them because I didn't think it did

cgwalters: Yeah you're right, it doesn't :cry:

cgwalters: OK I skimmed through the systemd sources here https://github.com/systemd/systemd/blob/361bf5fb770d4180eb5407f592a43c8f288c750f/src/libsystemd/sd-event/sd-event.c#L1312

Since I was curious if this thread spawning was really a best practice.  There's levels of nontrivial logic there.

But, systemd goes to great lengths to minimize its use of threads (particularly in pid1).  See e.g. https://sourceware.org/bugzilla/show_bug.cgi?id=26371

But in Rust land threads are the default and we have memory safety etc.

Would it work to just have a thread calling `waitpid(-1, ...)` in a loop?  That would seem to be dramatically simpler.

haircommander: good idea, this is a couple of lines simpler

conmon-rs/server/src/child_reaper.rs
haircommander: I like that

conmon-rs/server/src/child_reaper.rs
haircommander: apparently not

conmon-rs/server/src/child_reaper.rs
haircommander: one day..

conmon-rs/server/src/lib.rs
haircommander: pdeathsignal just kills conmon-rs and I'm not sure why. maybe something interacting with tokio?

conmon-rs/server/src/child_reaper.rs
mrunalp: https://github.com/tokio-rs/tokio/issues/4016

haircommander:
test failures were similar to https://github.com/kata-containers/kata-containers/issues/1826 -- thanks kata folks for an inspiration for the fix

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,156 @@
...
+                // we don't call waitpid until the runtime process has finished
+                // Inspired by
+                // https://github.com/kata-containers/kata-containers/blob/828a304883a4/src/agent/src/signal.rs#L27
+                let wait_status = {
+                    let wait_lock = Arc::clone(&wait_lock);
+                    let _lock = wait_lock.lock();
+                    waitpid(Pid::from_raw(-1), None)
+                };
+
+                let grandchildren = Arc::clone(&grandchildren);
```
saschagrunert, line 29: This additional clone can be removed.
```suggestion
```

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,156 @@
...
+        let grandchildren = Arc::clone(&self.grandchildren);
+        let wait_lock = Arc::clone(&self.wait_lock);
+        thread::spawn(move || {
+            loop {
+                // To prevent an error: "No child processes (os error 10)" when spawning new runtime processes,
+                // we don't call waitpid until the runtime process has finished
+                // Inspired by
+                // https://github.com/kata-containers/kata-containers/blob/828a304883a4/src/agent/src/signal.rs#L27
+                let wait_status = {
+                    let wait_lock = Arc::clone(&wait_lock);
```
saschagrunert, line 24: I think we can remove this clone:
```suggestion
```

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        // To prevent an error: "No child processes (os error 10)",
+        // we prevent our reaper thread from calling waitpid until this child has completed.
+        // Inspired by
+        // https://github.com/kata-containers/kata-containers/blob/828a304883a4/src/agent/src/signal.rs#L27
+        let wait_lock = Arc::clone(&self.wait_lock);
+        let _lock = wait_lock.lock().unwrap();
+
+        cmd.spawn()?.wait()
+    }
```
saschagrunert, line 84: I think this can be changed to avoid the `unwrap`. Also, I'm not sure if we really need the `let wait_lock = Arc::clone(&self.wait_lock);`:

```diff
diff --git a/conmon-rs/server/src/child_reaper.rs b/conmon-rs/server/src/child_reaper.rs
index d44756c9..4264ea7f 100644
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -1,6 +1,6 @@
 //! Child process reaping and management.
 
-use anyhow::{bail, Result};
+use anyhow::{bail, format_err, Context, Result};
 use getset::Getters;
 use log::{debug, error};
 use nix::sys::signal::{kill, Signal};
@@ -60,11 +60,7 @@ impl ChildReaper {
         Ok(())
     }
 
-    pub fn create_child<P, I, S>(
-        &self,
-        cmd: P,
-        args: I,
-    ) -> Result<std::process::ExitStatus, std::io::Error>
+    pub fn create_child<P, I, S>(&self, cmd: P, args: I) -> Result<std::process::ExitStatus>
     where
         P: AsRef<std::ffi::OsStr>,
         I: IntoIterator<Item = S>,
@@ -77,10 +73,12 @@ impl ChildReaper {
         // we prevent our reaper thread from calling waitpid until this child has completed.
         // Inspired by
         // https://github.com/kata-containers/kata-containers/blob/828a304883a4/src/agent/src/signal.rs#L27
-        let wait_lock = Arc::clone(&self.wait_lock);
-        let _lock = wait_lock.lock().unwrap();
+        let _ = &self
+            .wait_lock
+            .lock()
+            .map_err(|e| format_err!("take wait lock: {}", e))?;
 
-        cmd.spawn()?.wait()
+        Ok(cmd.spawn()?.wait().context("wait for child")?)
     }
 
     pub fn watch_grandchild(&self, id: String, pid: i32, exit_paths: Vec<PathBuf>) -> Result<()> {
```

conmon-rs/server/src/child_reaper.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    fn forget_grandchild(
+        locked_map: Arc<Mutex<HashMap<i32, ReapableChild>>>,
+        grandchild_pid: Pid,
+        exit_status: i32,
+    ) -> Result<()> {
+        debug!("caught signal for pid {}", grandchild_pid);
+
+        let mut map = locked_map.lock().unwrap();
```
saschagrunert, line 104: Let's do not unwrap:
```suggestion
        let mut map = locked_grandchildren
            .lock()
            .map_err(|e| format_err!("lock grandchildren: {}", e))?;
```

conmon-rs/server/src/child_reaper.rs
haircommander: if I don't have it I get 
```
error[E0382]: use of moved value: `grandchildren`
  --> conmon-rs/server/src/child_reaper.rs:23:43
   |
23 | ...                   thread::spawn(move || {
   |                                     ^^^^^^^ value moved into closure here, in previous iteration of loop
24 | ...                       if let Err(e) = Self::forget_grandchild(grandchildren, grandchild_pid, exit_status) {
   |                                                                   ------------- use occurs due to use in closure
   |
   = note: move occurs because `grandchildren` has type `Arc<std::sync::Mutex<HashMap<i32, ReapableChild>>>`, which does not implement the `Copy` trait

```

saschagrunert (approved):

PR #88, opened 2021-12-07
Add systemd logger

saschagrunert:
We disconnect the stdout/err streams from conmon at a certain point of
execution. Therefore it would make more sense to log to a dedicated
driver instead of those streams. We now add a `--log-driver` option as
well as a default systemd logger.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,6 +94,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "block-buffer"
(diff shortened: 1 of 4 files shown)


haircommander:
hm where did this error come from
```
time="2021-12-07T11:36:05Z" level=warning msg="unable to get oom kill count" error="no directory specified for memory.oom_control"
```

saschagrunert:
Rebased to make the tests green.

haircommander:
One more nit, can we have a None driver that prints to stdout and use it in most of the tests? it's useful to see the output IMO

haircommander:
that can be in a follow up if you'd like too

saschagrunert:
Sure, added a stdout logger as well :+1:

haircommander:
LGTM

PR #164, opened 2022-01-19
Pin clap to v3.0.12

saschagrunert:
We do not have to rely on the master build any more.


saschagrunert:
@haircommander rebased, PTAL

PR #171, opened 2022-01-21
add child struct and map to store child state

rphillips:
My hackday PR based on Peter's suggestion to figure out storing the child state.

Might be a better to do this.


saschagrunert:
@rphillips do you mind giving this one a rebase?

saschagrunert (approved):
LGTM

PR #198, opened 2022-02-08
Make `path_vec_from_text_list` more idiomatic

saschagrunert:
Making the helper function a bit more idiomatic Rust.

### DIFF
--- a/conmon-rs/server/src/rpc.rs
+++ b/conmon-rs/server/src/rpc.rs
@@ -79,10 +79,5 @@ fn pidfile_from_params(params: &conmon::CreateContainerParams) -> capnp::Result<
 }
 
 fn path_vec_from_text_list(tl: capnp::text_list::Reader) -> Result<Vec<PathBuf>, capnp::Error> {
-    let mut v: Vec<PathBuf> = vec![];
-    for t in tl {
-        let t_str = t?.to_string();
-        v.push(PathBuf::from(t_str));
-    }
-    Ok(v)
+    tl.iter().map(|r| r.map(PathBuf::from)).collect()
 }


conmon-rs/server/src/rpc.rs
```diff
@@ -79,10 +79,10 @@ fn pidfile_from_params(params: &conmon::CreateContainerParams) -> capnp::Result<
...
-        let t_str = t?.to_string();
-        v.push(PathBuf::from(t_str));
-    }
-    Ok(v)
+    Ok(tl
+        .iter()
+        .collect::<Result<Vec<_>, _>>()?
+        .iter()
+        .map(PathBuf::from)
+        .collect::<Vec<_>>())
```
cgwalters, line 87: But now we're allocating two `Vec<>`.  It's actually much simpler than that, see:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6a4412865d6c34aa039422c20e120830

Basically just `tl.into_iter().map(|r| r.map(PathBuf::from)).collect()` should work.

saschagrunert: Even better, thanks!

conmon-rs/server/src/rpc.rs
```diff
@@ -79,10 +79,5 @@ fn pidfile_from_params(params: &conmon::CreateContainerParams) -> capnp::Result<
 }
 
 fn path_vec_from_text_list(tl: capnp::text_list::Reader) -> Result<Vec<PathBuf>, capnp::Error> {
-    let mut v: Vec<PathBuf> = vec![];
-    for t in tl {
-        let t_str = t?.to_string();
-        v.push(PathBuf::from(t_str));
-    }
-    Ok(v)
+    tl.iter().map(|r| r.map(PathBuf::from)).collect()
```
cgwalters, line 82: Random aside, but it's not just more idiomatic, it should be more efficient because
the iterator implements [ExactSizeIterator](https://docs.capnproto-rust.org/capnp/traits/struct.ListIter.html#impl-ExactSizeIterator) and the `map()` -> `collect()` chain knows to use that to allocate exactly the right sized `Vec` from the start.  One needs to do this manually with explicit iteration, i.e. use `Vec::with_capacity(tl.len()); at the start.

cgwalters (approved):

conmon-rs/server/src/rpc.rs
saschagrunert: I was hoping that! We may go through our code base once we have an MVP to catch other places for such optimizations.

haircommander:
LGTM this does look more idiomatic

PR #200, opened 2022-02-08
Add static release build artifact

saschagrunert:
This allows us to re-use the pre-built binary in other projects, like CRI-O.


haircommander:
LGTM

PR #204, opened 2022-02-10
Rename `conmon-server` to `conmon` (v3.0.0)

saschagrunert:
The proposal is that we use `conmon` for the server binary and bump the development version to v3.0.0.

### DIFF
--- a/conmon-rs/server/Cargo.toml
+++ b/conmon-rs/server/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
-name = "conmon-server"
-version = "0.1.0"
+name = "conmon"
+version = "3.0.0-dev"
 edition = "2018"
 
 [[bin]]
-name = "conmon-server"
+name = "conmon"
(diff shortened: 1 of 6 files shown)


conmon-rs/server/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
-name = "conmon-server"
-version = "0.1.0"
+name = "conmon"
+version = "3.0.0"
```
haircommander, line 3: are we able to add an rc or anything? we may need to differentiate between the actual final 3.0.0 vs the in-progress one

saschagrunert: Sure, added a `-dev`

rphillips (approved):

rphillips (changes requested):
-dev is missing?

rphillips (approved):

rphillips (approved):

PR #210, opened 2022-02-14
exec sync support

rphillips:
- [X] ~Create a capnproto API to stream stdout/stderr from conmon-rs~
- [X] actually implement exec
- [x] race for some reason
- [x] migrate to tokio spawn
- [x] refactor the pid watcher

### DIFF
--- a/conmon-rs/common/proto/conmon.capnp
+++ b/conmon-rs/common/proto/conmon.capnp
@@ -1,6 +1,8 @@
 @0xffaaf7385bc4adad;
 
 interface Conmon {
+    ###############################################
+    # Version
     struct VersionResponse {
         version @0 :Text;
         tag @1: Text;
@@ -11,6 +13,8 @@ interface Conmon {
 
     version @0 () -> (response: VersionResponse);
 
+    ###############################################
+    # CreateContainer
     struct CreateContainerRequest {
         id @0 :Text;
         bundlePath @1 :Text;
@@ -23,4 +27,20 @@ interface Conmon {
     }
 
     createContainer @1 (request: CreateContainerRequest) -> (response: CreateContainerResponse);
+
+    ###############################################
+    # ExecSync
+    struct ExecSyncContainerRequest {
+        id @0 :Text;
+        timeout  @1 :Int32;
+        command @2 :List(Text);
+    }
+
+    struct ExecSyncContainerResponse {
+        exitCode @0 :Int32;
+        stdout @1 :Text;
+        stderr @2 :Text;
+    }
+
+    execSyncContainer @2 (request: ExecSyncContainerRequest) -> (response: ExecSyncContainerResponse);
 }
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -69,7 +69,7 @@ impl ChildReaper {
         let cleanup_grandchildren = locked_grandchildren.clone();
         let pid = child.pid;
         tokio::task::spawn(async move {
-            killed_channel.await.expect("no error on channel");
+            killed_channel.await.expect("error on channel");
             if let Err(e) = Self::forget_grandchild(&cleanup_grandchildren, pid) {
                 error!("error forgetting grandchild {}", e);
             }
@@ -129,6 +129,8 @@ impl ReapableChild {
                     }
                 }
                 Err(err) => {
+                    // TODO perhaps writing to the exit file anyway?
+                    // TODO maybe retry the waitpid?
                     if err != nix::errno::Errno::ECHILD {
                         error!("caught error in reading for sigchld {}", err);
                     }
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -1,169 +1,148 @@
 //! Child process reaping and management.
-
 use crate::child::Child;
-use anyhow::{bail, format_err, Result};
-use getset::Getters;
+use crate::console::Console;
+use anyhow::{format_err, Context, Result};
 use log::{debug, error};
+use multimap::MultiMap;
 use nix::sys::signal::{kill, Signal};
 use nix::sys::wait::{waitpid, WaitStatus};
 use nix::unistd::Pid;
 use std::path::PathBuf;
-use std::thread;
-use std::{collections::HashMap, fs::File, io::Write, sync::Arc, sync::Mutex};
+use std::sync::Mutex;
+use std::{fs::File, io::Write, sync::Arc};
+
+#[derive(Debug, Default)]
+pub struct ChildReaper {
+    grandchildren: Arc<Mutex<MultiMap<String, ReapableChild>>>,
+}
 
 impl ChildReaper {
-    pub fn init_reaper(&self) -> Result<()> {
-        let grandchildren = Arc::clone(&self.grandchildren);
-        let wait_lock = Arc::clone(&self.wait_lock);
-        thread::spawn(move || {
-            loop {
-                // To prevent an error: "No child processes (os error 10)" when spawning new runtime processes,
-                // we don't call waitpid until the runtime process has finished
-                // Inspired by
-                // https://github.com/kata-containers/kata-containers/blob/828a304883a4/src/agent/src/signal.rs#L27
-                let wait_status = {
-                    let _lock = wait_lock.lock();
-                    waitpid(Pid::from_raw(-1), None)
-                };
-
-                match wait_status {
-                    Ok(status) => {
-                        if let WaitStatus::Exited(grandchild_pid, exit_status) = status {
-                            let grandchildren = Arc::clone(&grandchildren);
-                            thread::spawn(move || {
-                                if let Err(e) = Self::forget_grandchild(
-                                    grandchildren,
-                                    grandchild_pid,
-                                    exit_status,
-                                ) {
-                                    error!(
-                                        "Failed to reap grandchild for pid {}: {}",
-                                        grandchild_pid, e
-                                    );
-                                }
-                            });
-                        }
-                    }
-
-                    Err(err) => {
-                        // TODO FIXME this busy loops right now while there are no grandchildren.
-                        // There should be a broadcast mechanism so we only run this loop
-                        // while there are grandchildren
-                        if err != nix::errno::Errno::ECHILD {
-                            error!("caught error in reading for sigchld {}", err);
-                        }
-                    }
-                }
-            }
-        });
-        Ok(())
+    pub fn get(&self, id: String) -> Result<ReapableChild> {
+        let locked_grandchildren = Arc::clone(&self.grandchildren);
+        let lock = locked_grandchildren.lock().unwrap();
+        let r = lock.get(&id).context("")?.clone();
+        drop(lock);
+        Ok(r)
     }
 
-    pub fn create_child<P, I, S>(&self, cmd: P, args: I) -> Result<std::process::ExitStatus>
+    pub async fn create_child<P, I, S>(
+        &self,
+        cmd: P,
+        args: I,
+        console: Option<Console>,
+        pidfile: PathBuf,
+    ) -> Result<i32>
     where
         P: AsRef<std::ffi::OsStr>,
         I: IntoIterator<Item = S>,
         S: AsRef<std::ffi::OsStr>,
     {
-        let mut cmd = std::process::Command::new(cmd);
+        let mut cmd = tokio::process::Command::new(cmd);
         cmd.args(args);
-
-        // To prevent an error: "No child processes (os error 10)",
-        // we prevent our reaper thread from calling waitpid until this child has completed.
-        // Inspired by
-        // https://github.com/kata-containers/kata-containers/blob/828a304883a4/src/agent/src/signal.rs#L27
-        let wait_lock = Arc::clone(&self.wait_lock);
-        let _lock = wait_lock
-            .lock()
-            .map_err(|e| format_err!("lock waitlock: {}", e))?;
-
         cmd.spawn()
-            .map_err(|e| format_err!("spawn child process: {}", e))?
+            .context("spawn child process: {}")?
             .wait()
-            .map_err(|e| format_err!("wait for child process: {}", e))
+            .await?;
+
+        if let Some(console) = console {
+            let _ = console
+                .wait_connected()
+                .context("wait for console socket connection");
+        }
+
+        let grandchild_pid = tokio::fs::read_to_string(pidfile)
+            .await?
+            .parse::<i32>()
+            .context("grandchild pid parse error")?;
+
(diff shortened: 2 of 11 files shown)


conmon-rs/common/proto/conmon.capnp
```diff
@@ -19,4 +23,33 @@ interface Conmon {
...
+
+    enum ExecChunkType {
+        stdout @0;
+        stderr @1;
+    }
+
+    interface ExecCallback {
+        sendChunk @0 (type :ExecChunkType, chunk :Data);
+        done @1 ();
+    }
```
saschagrunert, line 52: I was thinking if it would make sense to merge `done` into `sendChunk` and pass the exit code as another parameter. But I guess having two methods is fine as well.

haircommander:
while this is a good start, I think it may be better to focus on ExecSync requests--ones that wait for the entirety of the output from the exec request. this is currently what CRI-O uses conmon for, and it will be simpler than a stream

conmon-rs/server/src/rpc.rs
```diff
@@ -69,6 +69,34 @@ impl conmon::Server for Server {
...
+            let child_reaper = Arc::clone(self.reaper());
+            let command = self.generate_exec_sync_args(&params).unwrap();
+            let (exit_status, stdout, stderr) =
+                pry_err!(child_reaper.exec_sync(self.config().runtime(), command));
+            let mut resp = results.get().init_response();
+            resp.set_exit_code(exit_status.code().unwrap());
+            resp.set_stdout(stdout.as_str());
+            resp.set_stderr(stderr.as_str());
+        } else {
+            // container does not exist
```
haircommander, line 93: in this situation we should error. I would reverse the blocks and return with an error if None here

conmon-rs/server/src/child_reaper.rs
```diff
@@ -59,40 +60,77 @@ impl ChildReaper {
...
+        let delay = tokio::time::sleep(tokio::time::Duration::from_secs(timeout as u64));
+        tokio::pin!(delay);
+        tokio::select! {
+            _ = &mut delay => {
+                debug!("timeout");
+                return Err(format_err!("timeout error"))
+            },
+            status = child.wait_with_output() => {
+                 let output = status.expect("blah");
+                 let stdout = String::from_utf8(output.stdout)?;
```
cgwalters, line 101: I think we don't want to error out if the child output invalid UTF-8 somehow.  I'd use [from_utf8_lossy](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy) instead.

(And what's better, we avoid duplicating the buffer in the happy path)

However, since this function returns owned values, we probably want to change it to return the raw `Vec<u8>`, and do the utf8 bits in the callers.

conmon-rs/server/src/rpc.rs
```diff
@@ -48,29 +49,80 @@ impl conmon::Server for Server {
...
+        let runtime = self.config.runtime().clone();
+        debug!(
+            "Got exec sync container request for id {} with timeout {} : {}",
+            id,
+            timeout,
+            command.join(" ")
+        );
+        let children = self.children.read().unwrap();
+        let child = children.get(id);
+        if child.is_none() {
```
cgwalters, line 108: There's a bit more verbose pattern for this
```
let child = if let Some(c) = child {
    c
} else {
    let mut resp = results.get().init_response();
    resp.set_exit_code(-1);
    return Promise::ok(());
};
```
And then below we can drop the `unwrap()`.

In the future https://rust-lang.github.io/rfcs/3137-let-else.html will help this.

conmon-rs/server/src/rpc.rs
```diff
@@ -48,29 +49,80 @@ impl conmon::Server for Server {
...
+        debug!("found child with id {}", child.unwrap().id);
+        let child_reaper = Arc::clone(self.reaper());
+        Promise::from_future(async move {
+            let (exit_status, stdout, stderr) = child_reaper
+                .exec_sync(&runtime, command, timeout)
+                .await
+                .map_err(|e| Error::new(ErrorKind::Other, format!("Error {}", e)))?;
+            let mut resp = results.get().init_response();
+            resp.set_exit_code(exit_status.code().unwrap());
+            resp.set_stdout(stdout.as_str());
```
cgwalters, line 122: If `exec_sync` returned `Vec<u8>` this is where we could do `from_utf8_lossy()`.

conmon-rs/server/src/rpc.rs
rphillips: thanks, incorporated this in!

conmon-rs/server/src/rpc.rs
rphillips: thanks. fixed :)

conmon-rs/server/src/child_reaper.rs
```diff
@@ -7,133 +7,88 @@ use log::{debug, error};
...
+        args: Vec<String>,
+        timeout: i32,
+    ) -> Result<Output> {
+        let child = tokio::process::Command::new(command)
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .kill_on_drop(true)
+            .args(args)
+            .spawn()
+            .map_err(|e| format_err!("spawn child process: {}", e))?;
```
saschagrunert, line 42: Let's replace those with the `anyhow::Context`:
```suggestion
            .context("spawn child process")?;
```

conmon-rs/server/src/child_reaper.rs
```diff
@@ -7,133 +7,88 @@ use log::{debug, error};
...
+
+        let delay = tokio::time::sleep(tokio::time::Duration::from_secs(timeout as u64));
+        tokio::pin!(delay);
+        tokio::select! {
+            _ = &mut delay => {
+                debug!("timeout");
+                return Err(format_err!("timeout error"))
```
saschagrunert, line 49:
```suggestion
                bail!("timeout error")
```

conmon-rs/server/src/child_reaper.rs
```diff
@@ -7,133 +7,88 @@ use log::{debug, error};
...
+            },
+            status = child.wait_with_output() => {
+                 let output = status.expect("status expected");
+                 Ok(output)
```
saschagrunert, line 53:
```suggestion
                 Ok(status?)
```

conmon-rs/server/src/lib.rs
```diff
@@ -236,4 +246,28 @@ impl Server {
...
+    fn generate_exec_sync_args(
+        &self,
+        params: &conmon::ExecSyncContainerParams,
+    ) -> Result<Vec<String>> {
+        let req = params.get()?.get_request()?;
+        let id = req.get_id()?.to_string();
+        let command = req.get_command()?;
+        let runtime_root = self.config().runtime_root();
+
+        let mut args: Vec<String> = vec![];
```
saschagrunert, line 260: nit:
```suggestion
        let mut args = vec![];
```

conmon-rs/server/src/lib.rs
```diff
@@ -236,4 +246,28 @@ impl Server {
...
+        if let Some(rr) = runtime_root {
+            args.push(format!("--root={}", rr.display()));
+        }
+        args.push("exec".to_string());
+        args.push(id);
+        for value in command.iter() {
+            args.push(value.unwrap().to_string());
```
saschagrunert, line 267:
```suggestion
            args.push(value?.to_string());
```

conmon-rs/server/src/lib.rs
rphillips: fixed!

conmon-rs/server/src/child_reaper.rs
rphillips: fixed!

conmon-rs/server/src/child_reaper.rs
rphillips: fixed!

conmon-rs/server/src/child_reaper.rs
rphillips: fixed!

rphillips:
Need to figure out the stdout/stderr streams... and timeouts.

Transitioned to using "-d" for detach on the exec.

conmon-rs/server/src/child_reaper.rs
```diff
@@ -1,169 +1,148 @@
...
-                }
-            }
-        });
-        Ok(())
+impl ChildReaper {
+    pub fn get(&self, id: String) -> Result<ReapableChild> {
+        let locked_grandchildren = Arc::clone(&self.grandchildren);
+        let lock = locked_grandchildren.lock().unwrap();
+        let r = lock.get(&id).context("")?.clone();
+        drop(lock);
```
haircommander, line 24: do we need to manually call drop here? I would expect it to go out of scope after the Ok

haircommander:
could you separate the commits to be smaller pieces of work?

conmon-rs/server/src/child_reaper.rs
rphillips: yes, the drop needs to be called to allow the borrow checker to succeed.

rphillips:
This is ready for review.

There are a couple pending TODOs. The console library needs a refactor to propagate up the stdout/stderr of the grandchild. The secondary TODO is to return the grandchild's exit code.

I think we should merge this and then tackle the TODOs in parallel.

conmon-rs/common/proto/conmon.capnp
```diff
@@ -23,4 +27,20 @@ interface Conmon {
     }
 
     createContainer @1 (request: CreateContainerRequest) -> (response: CreateContainerResponse);
-}
+
```
haircommander, line 30: i think these changes should go in the second commit

conmon-rs/server/src/lib.rs
```diff
@@ -236,4 +235,36 @@ impl Server {
...
+        }
+        args.push(format!("--pid-file={}", pidfile.display()));
+        args.push(id);
+        for value in command.iter() {
+            args.push(value?.to_string());
+        }
+
+        debug!("Exec args {:?}", args.join(" "));
+        Ok(args)
+    }
```
haircommander, line 269: as well as these (and any specifically exec related pieces)

conmon-rs/server/src/child_reaper.rs
```diff
@@ -1,169 +1,148 @@
...
+                    if err != nix::errno::Errno::ECHILD {
+                        error!("caught error in reading for sigchld {}", err);
+                    }
+                }
+            };
+            if tx.send(()).is_err() {
+                error!("the receiver dropped the watch")
+            }
+        });
+        rx
```
haircommander, line 141: whereas the refactoring of child reaping and pid watching should be in a separate commit describing the reasoning

conmon-rs/server/src/rpc.rs
```diff
@@ -48,29 +48,82 @@ impl conmon::Server for Server {
...
+            // TODO FIXME why convert?
+            results
+                .get()
+                .init_response()
+                .set_container_pid(grandchild_pid as u32);
+            Ok(())
+        })
+    }
+
+    fn exec_sync_container(
```
haircommander, line 76: same here--second commit adding exec please

pkg/client/client.go
```diff
@@ -166,6 +166,68 @@ func (c *ConmonClient) CreateContainer(ctx context.Context, cfg *CreateContainer
 	return response.ContainerPid(), nil
 }
 
+type ExecContainerResult struct {
```
haircommander, line 169: client update could even go in another separate commit--but that's up to you. it could also go with the exec commit if you'd like

haircommander (changes requested):
:tada: great work! I have described what I'm looking for in commit structure a bit more thoroughly. My first glance LGTM, once this is reorganized i'll do another more in-depth review

conmon-rs/common/proto/conmon.capnp
rphillips: hmm. the second commit is just the regeneration of the files.

conmon-rs/server/src/child_reaper.rs
```diff
@@ -1,169 +1,148 @@
...
+                Ok(status) => {
+                    if let WaitStatus::Exited(_, exit_status) = status {
+                        let _ = write_to_exit_paths(exit_status, &exit_paths);
+                    }
+                }
+                Err(err) => {
+                    if err != nix::errno::Errno::ECHILD {
+                        error!("caught error in reading for sigchld {}", err);
+                    }
+                }
```
haircommander, line 135: nit: we may want a TODO here. Part of me thinks we may want to write to the exit file anyway so the caller knows the container exited.

nit2: it's possible we want to retry in this case. we'll have to stress conmon more to figure it out

conmon-rs/server/src/child_reaper.rs
```diff
@@ -1,169 +1,148 @@
...
-        if let Some(old) = map.insert(child.pid, reapable_grandchild) {
-            bail!("Repeat PID for container {} found", old.id);
-        }
+        let reapable_grandchild = ReapableChild::from_child(&child);
+        let killed_channel = reapable_grandchild.watch();
+        map.insert(child.id, reapable_grandchild);
+        let cleanup_grandchildren = locked_grandchildren.clone();
+        let pid = child.pid;
+        tokio::task::spawn(async move {
+            killed_channel.await.expect("no error on channel");
```
haircommander, line 72: wouldn't calling expect("no error on channel") mean we print an error saying "no error on channel" in cases where there was an error on the channel?

conmon-rs/server/src/child_reaper.rs
```diff
@@ -1,169 +1,148 @@
...
+            if let Err(e) = Self::forget_grandchild(&cleanup_grandchildren, pid) {
```
haircommander, line 73: also, if we're waiting on the channel here to forget the granchild, why don't we just call waitpid within this task, rather than needing to sync between two tasks?

haircommander (approved):
A couple of more questions. I don't think any are blocking, but I wanted to get them out there. Being added as TODOs would be fine with me

Otherwise, LGTM

conmon-rs/server/src/child_reaper.rs
rphillips: hah. yeah... I'll fix that :)

conmon-rs/server/src/child_reaper.rs
rphillips: since this is the 'sync' function I am thinking we want to wait on the async task to complete. Ideally, we want less code and usually the 'sync' version is just a block on the async version.

conmon-rs/server/src/child_reaper.rs
rphillips: fixed.

saschagrunert (approved):
Great work, let's follow-up on the open discussion items and iterate.

PR #211, opened 2022-02-15
Make `conmon --version` CRI-O compatible

saschagrunert:
CRI-O's conmon version parsing fails right now there:
https://github.com/cri-o/cri-o/blob/2e81aea6aa5d7a2fab42c07cfbe7a3f6299cea23/internal/config/conmonmgr/conmonmgr.go#L35-L37

To fix that, we now prepend the crate name to the version output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -245,7 +245,7 @@ dependencies = [
 
 [[package]]
 name = "conmon"
-version = "3.0.0"
+version = "3.0.0-dev"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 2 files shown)


haircommander:
as a follow up, we should add this function to the library here so callers can use it directly!

there's also [probe conmon](https://github.com/containers/common/blob/5a035772e4a299d6bdd95094a06af09aee452e8d/pkg/config/default.go#L384) in c/common

haircommander:
LGTM

PR #212, opened 2022-02-16
Add version fields to RPC

saschagrunert:
We now also make the `Version` module partly public for external usage.

Follow-up of https://github.com/containers/conmon-rs/pull/211


haircommander:
LGTM

PR #227, opened 2022-02-23
Don't use `.unwrap()` any more

saschagrunert:
We now handle error paths which used `unwrap` before.

### DIFF
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -1,6 +1,5 @@
 //! Child process reaping and management.
-use crate::child::Child;
-use crate::console::Console;
+use crate::{child::Child, console::Console};
 use anyhow::{format_err, Context, Result};
 use log::{debug, error};
 use multimap::MultiMap;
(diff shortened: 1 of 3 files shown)


rphillips:
much needed. thanks!

conmon-rs/server/src/child_reaper.rs
```diff
@@ -16,10 +15,16 @@ pub struct ChildReaper {
...
+    ($x:expr) => {
+        $x.lock().map_err(|e| format_err!("{:#}", e))?
+    };
+}
+
 impl ChildReaper {
     pub fn get(&self, id: String) -> Result<ReapableChild> {
         let locked_grandchildren = Arc::clone(&self.grandchildren);
-        let lock = locked_grandchildren.lock().unwrap();
+        let lock = lock!(locked_grandchildren);
```
cgwalters, line 27: The only reason `lock()` can fail is if *another* thread holding the lock panic'd.

I think it's actually reasonable in this scenario to make the panic propagate, rather than turn into a regular error.

Now in some codebases, any `unwrap()/expect()` calls need e.g. a `SAFETY: ` comment.

But just for reference, in e.g. the rust compiler itself, there's a fair bit of usage of `.lock().unwrap()`: https://cs.github.com/?scopeName=All+repos&scope=&q=repo%3Arust-lang%2Frust+%2F%5C.lock.*unwrap%2F

PR #232, opened 2022-02-24
Make execsync output work

saschagrunert:
We now skip the IOStreams for execsync and rely on a dedicated read loop for now.


conmon-rs/server/src/rpc.rs
```diff
@@ -105,19 +115,30 @@ impl conmon::Server for Server {
...
+                    let mut stdout = String::new();
+                    for msg in console.message_rx() {
+                        match msg {
+                            Message::Data(s) => stdout.push_str(&s),
+                            Message::Done => break,
+                        }
+                    }
+
+                    resp.set_stdout(&stdout);
                     resp.set_exit_code(0);
```
saschagrunert, line 138: @rphillips @haircommander have you discussed stderr handling when using the runc file descriptor passing? I thought we could not distinguish stdout from stderr in that case.

conmon-rs/server/src/console.rs
```diff
@@ -29,6 +32,15 @@ pub struct Console {
...
 
     connected_rx: Receiver<()>,
+
+    #[getset(get = "pub")]
+    message_rx: Receiver<Message>,
+}
+
+#[derive(Debug)]
+pub enum Message {
+    Data(String),
```
saschagrunert, line 42: Copying around strings is probably not the most efficient way.

conmon-rs/server/src/rpc.rs
rphillips: Ah, I wasn't aware of that... We should remove the stderr components.

saschagrunert: Good, I'll prepare that.

saschagrunert:
Ready for review. PTAL @rphillips @haircommander

rphillips (approved):

conmon-rs/server/src/rpc.rs
```diff
@@ -105,19 +115,30 @@ impl conmon::Server for Server {
...
-                    let _ = child_reaper.watch_grandchild(child);
+                    capnp_err!(child_reaper.watch_grandchild(child))?;
+
                     // TODO return the grandchild exit code
-                    // TODO return the stdout and stderr of the child
+
+                    let mut stdio = vec![];
+                    for msg in console.message_rx() {
+                        match msg {
+                            Message::Data(s) => stdio.extend(s),
```
haircommander, line 132: are we at risk of DOSing the server if this exec request takes a long time?

conmon-rs/common/proto/conmon.capnp
```diff
@@ -38,9 +38,8 @@ interface Conmon {
 
     struct ExecSyncContainerResponse {
         exitCode @0 :Int32;
-        stdout @1 :Text;
-        stderr @2 :Text;
+        stdio @1 :Data;
```
haircommander, line 41: I think we still need a differentiation between the two, even if a tty request only sends in stdout

saschagrunert: Good, added it back.

conmon-rs/server/src/rpc.rs
saschagrunert: I took the time out into account, this should prevent such attacks.

mrunalp:
I just caught up reviewing the overall state of conmon rs today :) I have a question about the exec sync implementation: Why can't we use tokio process for it directly? That captures the output for us nicely.

saschagrunert:
> I just caught up reviewing the overall state of conmon-rs today :)

Awesome, thank you! :blush: 

> I have a question about the exec sync implementation: Why can't we use tokio process for it directly? That captures the output for us nicely. It has a kill_on_drop option that we can combine with a timeout on a select.

Yeah this could probably work. @rphillips @haircommander do we wanna re-sync on the execsync architecture?

rphillips:
@mrunalp I tried that approach... The problem is that conmon (today) detaches the exec sync call so the exec's grandchild can be reaped during a container shutdown.

haircommander:
> @mrunalp I tried that approach... The problem is that conmon (today) detaches the exec sync call so the exec's grandchild can be reaped during a container shutdown.

also for a exec sync with terminal we need to detach. doing so in both cases simplifies the code I think

rphillips:
@saschagrunert I merged the pid PR... this pr will need a rebase.

rphillips (approved):

saschagrunert:
Thank you for the clarification @rphillips @haircommander, I released the PR back on top of `main`.

rphillips (approved):

PR #234, opened 2022-02-28
Use nix::errno for set_child_subreaper

saschagrunert:
This removes the error conversion by directly mapping it from the returned `i32`.


haircommander:
LGTM

PR #239, opened 2022-03-03
Fix error handling paths

saschagrunert:
Some errors can actually be returned or mapped by using the convenience
macro. We also fix a test name for the integration tests.

### DIFF
--- a/conmon-rs/server/src/console.rs
+++ b/conmon-rs/server/src/console.rs
@@ -274,7 +274,7 @@ mod tests {
 
         // Write to the slave
         let mut file = unsafe { fs::File::from_raw_fd(res.slave) };
-        let _ = file.write(b"test").await?;
+        file.write_all(b"test").await?;
 
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


rphillips (approved):

conmon-rs/server/src/console.rs
```diff
@@ -274,7 +274,7 @@ mod tests {
 
         // Write to the slave
         let mut file = unsafe { fs::File::from_raw_fd(res.slave) };
-        let _ = file.write(b"test").await?;
+        file.write(b"test").await?;
```
cgwalters, line 277: This one is more subtle.  It should actually be `write_all()`.  While it doesn't usually happen for local files in the normal path, one almost always wants to use `write_all` unless dealing with a network socket.

saschagrunert: Yep, switched to `write_all`.

rphillips (approved):

PR #255, opened 2022-03-15
Use busybox binary from repo

saschagrunert:
The integration tests recently flaked because of the ad-hoc download of
the busybox binary. To avoid such failures we now commit the binary into
the repo and adapt the test paths accordingly.


haircommander:
good idea LGTM

cgwalters:
Committing binaries to git is pretty eww.  Not to mention it won't work on non-x86 architectures.  I'd start with adding some retries on downloads.

saschagrunert:
Hm, I think I'll keep it as-is for now. The binary only has a few KB so I thought it would not hurt to have it in the repo.

PR #268, opened 2022-03-25
client rework and exec sync fix

haircommander:
This PR has three fixes needed for the cri-o integration. With this, many cri-o tests pass with [this](https://github.com/haircommander/cri-o/tree/runtime-pod) branch of cri-o!


conmon-rs/server/src/child_reaper.rs
```diff
@@ -125,22 +124,27 @@ impl ReapableChild {
...
+        let (exit_tx, exit_rx) = oneshot::channel();
 
         task::spawn_blocking(move || {
             let wait_status = waitpid(Pid::from_raw(pid as pid_t), None);
             match wait_status {
                 Ok(status) => {
                     if let WaitStatus::Exited(_, exit_code) = status {
-                        Self::write_to_exit_paths(exit_code, &exit_paths)?;
-
+                        // HACK: keep this before write_to_exit_paths.
```
haircommander, line 138: @saschagrunert @rphillips @cgwalters any suggestions here? We probably need a different structure...

pkg/client/client.go
```diff
@@ -70,28 +59,107 @@ func New(config *ConmonServerConfig) (_ *ConmonClient, retErr error) {
...
+	if config.Stdin != nil {
+		cmd.Stdin = os.Stdin
+	}
+	if config.Stdout != nil {
+		cmd.Stdout = os.Stdout
+	}
+	if config.Stderr != nil {
+		cmd.Stderr = os.Stderr
+	}
+	return cmd.Run()
```
rphillips, line 96: should this be Start() so it doesn't block?

pkg/client/client.go
```diff
@@ -70,28 +59,107 @@ func New(config *ConmonServerConfig) (_ *ConmonClient, retErr error) {
...
+	}, nil
+}
+
+func (c *ConmonClient) StartServer(config *ConmonServerConfig) error {
+	entrypoint, args, err := c.toArgs(config)
+	if err != nil {
+		return err
+	}
+	cmd := exec.Command(entrypoint, args...)
+	if config.Stdin != nil {
```
rphillips, line 87: This config stdio block was in the original code, but it appears like it should be assigning config.Stdin and friends... perhaps defaulting the os.Std* first

conmon-rs/server/src/child_reaper.rs
```diff
@@ -151,12 +155,12 @@ impl ReapableChild {
...
                 }
             };
-            if tx.send(()).is_err() {
+            if killed_tx.send(()).is_err() {
                 bail!("the receiver dropped the watch")
             }
             Ok(())
         });
-        rx
+        (killed_rx, exit_rx)
```
saschagrunert, line 163: Can we rename the `killed` channel to something like `done`? Because it indicates that we actively killed a process which can cause confusion.

conmon-rs/server/src/child_reaper.rs
saschagrunert: I think the intention to write the exit paths first and then send to the channel was to indicate that we're done with the overall work. Now there is a chance for a race where someone pulled the exit code but we still have no exit paths written.

Hm, can we somehow use `killed_rx` to send the exit code by using a multi-consumer channel? 
https://docs.rs/crossbeam/0.8.1/crossbeam/channel/index.html

conmon-rs/server/src/child_reaper.rs
saschagrunert, line 163: Can we rename the `killed` channel to something like `done`? Because it indicates that we actively killed a process which can cause confusion.

conmon-rs/server/src/child_reaper.rs
haircommander: crossbeam and tokio don't play nice together https://github.com/tokio-rs/tokio/issues/2447

but we could achieve the same thing with tokio's broadcast  I think

pkg/client/client.go
haircommander: oh yeah this is definitely messed up haha

haircommander: fixed

conmon-rs/server/src/child_reaper.rs
haircommander: dropped the channel anyway

conmon-rs/server/src/child_reaper.rs
haircommander: done!

haircommander:
so exec sync still breaks some tests because it always creates a terminal, which adds extra characters that the crio tests don't know about . I think we need to refactor iostreams structure so that the data can be pulled from for exec sync (and eventually attach) in the same way console is. @saschagrunert if you find it convenient, could you look at this? If not I will tomorrow, but you're closer to the iostream code

saschagrunert:
> so exec sync still breaks some tests because it always creates a terminal, which adds extra characters that the crio tests don't know about . I think we need to refactor iostreams structure so that the data can be pulled from for exec sync (and eventually attach) in the same way console is. @saschagrunert if you find it convenient, could you look at this? If not I will tomorrow, but you're closer to the iostream code

I can look into this, but would prefer merging this PR before because I cannot push to your fork.
To fix the tests, let's run them as root:
```diff
diff --git a/Makefile b/Makefile
index 36411cc4..49786a6e 100644
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,9 @@ unit:
 	cargo test --bins --no-fail-fast
 
 integration: release # It needs to be release so we correctly test the RSS usage
-	CONMON_BINARY="$(MAKEFILE_PATH)target/release/$(BINARY)" RUNTIME_BINARY="$(RUNTIME_PATH)" go test -v pkg/client/*
+	export CONMON_BINARY="$(MAKEFILE_PATH)target/release/$(BINARY)" && \
+	export RUNTIME_BINARY="$(RUNTIME_PATH)" && \
+	sudo -E go test -count=1 -ginkgo.v -v pkg/client/*
 
 clean:
 	rm -rf target/
```

conmon-rs/server/src/child_reaper.rs
saschagrunert: Can we remove the `HACK` comment now?

conmon-rs/server/src/rpc.rs
```diff
@@ -121,7 +121,13 @@ impl conmon::Server for Server {
...
                     // register grandchild with server
                     let child = Child::new(id, grandchild_pid, child.exit_paths);
-                    let exit_tx = capnp_err!(child_reaper.watch_grandchild(child))?;
+
+                    let mut exit_tx = capnp_err!(child_reaper.watch_grandchild(child))?;
+
+                    // Immediately attempt to pull the exit code
+                    // this is to work around a limitation of oneshot that causes an error
+                    // if the reader reads after the sender is closed (goes out of scope).
+                    let ec = capnp_err!(exit_tx.recv().await)?;
```
saschagrunert, line 130: The timeout does not work anymore with this change, let's fix that by using the tokio timeout:
```diff
diff --git a/conmon-rs/server/src/rpc.rs b/conmon-rs/server/src/rpc.rs
index 3e90d9e2..a83e7366 100644
--- a/conmon-rs/server/src/rpc.rs
+++ b/conmon-rs/server/src/rpc.rs
@@ -91,12 +91,12 @@ impl conmon::Server for Server {
     ) -> Promise<(), capnp::Error> {
         let req = pry!(pry!(params.get()).get_request());
         let id = pry!(req.get_id()).to_string();
-        let timeout = req.get_timeout_sec();
+        let timeout_sec = req.get_timeout_sec();
         let pidfile = pry_err!(Console::temp_file_name("exec_sync", "pid"));
 
         debug!(
-            "Got exec sync container request for id {} with timeout {}",
-            id, timeout,
+            "Got exec sync container request for id {} with timeout of {} sec",
+            id, timeout_sec,
         );
 
         let console = pry_err!(Console::new());
@@ -127,14 +127,15 @@ impl conmon::Server for Server {
                     // Immediately attempt to pull the exit code
                     // this is to work around a limitation of oneshot that causes an error
                     // if the reader reads after the sender is closed (goes out of scope).
-                    let ec = capnp_err!(exit_tx.recv().await)?;
+                    let timeout = Duration::from_secs(timeout_sec);
+                    let ec = capnp_err!(capnp_err!(
+                        tokio::time::timeout(timeout, exit_tx.recv()).await
+                    )?)?;
 
                     let mut stdio = vec![];
                     loop {
-                        let msg = if timeout > 0 {
-                            capnp_err!(console
-                                .message_rx()
-                                .recv_timeout(Duration::from_secs(timeout)))?
+                        let msg = if timeout_sec > 0 {
+                            capnp_err!(console.message_rx().recv_timeout(timeout))?
                         } else {
                             capnp_err!(console.message_rx().recv())?
                         };
```

haircommander: ah we can undo the changes done previously, that was a workaround for the "sender closed" issue

pkg/client/client.go
haircommander: I think we want it to block. it double forks anyway so we won't block after the server is up

saschagrunert (approved):
LGTM, feel free to merge If you’re done

rphillips:
lgtm

haircommander:
let's merge this and I'll keep iterating, there's more where this came from :sunglasses:

PR #274, opened 2022-03-30
Use cri-o bucket for busybox test binary

saschagrunert:
To avoid more flaky tests we now use our own Google Cloud bucket for the
test binary rather than the official sources which often timeout.


haircommander:
LGTM

PR #275, opened 2022-03-30
Retry waitpid and always write exit paths

saschagrunert:
This fixes two `TODO`s by retrying `waitpid` on `EINTR`. Beside that we
now always try to write the exit paths in different contexts.


conmon-rs/server/src/child_reaper.rs
```diff
@@ -132,28 +132,40 @@ impl ReapableChild {
...
-                    if err != Errno::ECHILD {
-                        bail!("caught error in reading for sigchld {}", err);
+                    Err(err) if err != Errno::ECHILD => {
+                        Self::write_to_exit_paths(FAILED_EXIT_CODE, &exit_paths)
+                            .context("write to exit paths on waitpid err != ECHILD")?;
+                        bail!("caught error in reading for sigchld {}", err)
                     }
-                    if tx_clone.send(-3).is_err() {
-                        bail!("unable to send exit status")
+                    Err(err) if err != Errno::EINTR => {
```
haircommander, line 154: wouldn't this fail on ECHILD now?

conmon-rs/server/src/child_reaper.rs
```diff
@@ -132,28 +132,40 @@ impl ReapableChild {
...
-                }
-                Err(err) => {
-                    // TODO perhaps writing to the exit file anyway?
-                    // TODO maybe retry the waitpid?
-                    if err != Errno::ECHILD {
-                        bail!("caught error in reading for sigchld {}", err);
+                    Err(err) if err != Errno::ECHILD => {
+                        Self::write_to_exit_paths(FAILED_EXIT_CODE, &exit_paths)
+                            .context("write to exit paths on waitpid err != ECHILD")?;
+                        bail!("caught error in reading for sigchld {}", err)
```
haircommander, line 152: can we unify all of this by using this block to find an exit code, and then breaking out and writing/sending that exit code?

conmon-rs/server/src/child_reaper.rs
saschagrunert: Ooops this logic is incorrect, fixing it.

conmon-rs/server/src/child_reaper.rs
saschagrunert: Yes, changed the logic to refactor everything into a dedicated funciton.

conmon-rs/server/src/child_reaper.rs
```diff
@@ -132,31 +132,40 @@ impl ReapableChild {
         let tx_clone = exit_tx.clone();
 
         task::spawn_blocking(move || {
-            let wait_status = waitpid(Pid::from_raw(pid as pid_t), None);
-            match wait_status {
+            let exit_code = Self::wait_for_exit_code(pid)?;
```
haircommander, line 135: do we risk waiting forever on the exit_rx side if we bail here?

haircommander: this code doesn't change that behavior but I just thought of it

haircommander:
LGTM

haircommander:
LGTM

PR #284, opened 2022-04-01
client, server: add and use runtime dir instead of relying on temp files

haircommander:
This way, we know a conmon-rs's files are cleaned up after its runtime dir is

### DIFF
--- a/conmon-rs/server/src/config.rs
+++ b/conmon-rs/server/src/config.rs
@@ -53,16 +53,6 @@ pub struct Config {
 
     #[get = "pub"]
     #[clap(
-        env(concat!(prefix!(), "PIDFILE")),
-        long("conmon-pidfile"),
-        short('P'),
-        value_name("PIDFILE")
-    )]
-    /// PID file for the conmon server.
-    conmon_pidfile: Option<PathBuf>,
-
-    #[get = "pub"]
-    #[clap(
         default_value_if("version", None, Some("")),
         env(concat!(prefix!(), "RUNTIME")),
         long("runtime"),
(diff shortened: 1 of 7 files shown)


haircommander:
maybe a decent TODO: add a test making sure those consts are in sync. linking between rust and go isn't as trivial as go and c, so the approach of conmon didn't seem correct

conmon-rs/server/src/config.rs
```diff
@@ -155,4 +152,14 @@ impl Config {
 
         Ok(())
     }
+    pub fn socket(&self) -> PathBuf {
+        let mut path = self.runtime_dir().clone();
+        path.push(SOCKET);
+        path
```
saschagrunert, line 158:
```suggestion
        self.runtime_dir().join(SOCKET)
```

conmon-rs/server/src/config.rs
```diff
@@ -155,4 +152,14 @@ impl Config {
...
+    }
+    pub fn conmon_pidfile(&self) -> PathBuf {
+        let mut path = self.runtime_dir().clone();
+        path.push(PIDFILE);
+        path
```
saschagrunert, line 163:
```suggestion
        self.runtime_dir().join(PIDFILE)
```

conmon-rs/server/src/container_io.rs
```diff
@@ -1,6 +1,7 @@
 use crate::{streams::Streams, terminal::Terminal};
 use anyhow::{Context, Result};
 use crossbeam_channel::Sender;
+use std::path::Path;
 use std::sync::mpsc::Receiver;
```
saschagrunert, line 5: nit:
```suggestion
use std::{path::Path, sync::mpsc::Receiver};
```

haircommander:
updated as suggested!

saschagrunert (approved):

PR #289, opened 2022-04-04
rename conmon binary to conmonrs

rphillips:
Rename the binary to conmonrs, since it is not 100% compatible with conmon.


Makefile
```diff
@@ -1,7 +1,7 @@
 MAKEFILE_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 RUNTIME_PATH ?= "/usr/bin/runc"
 PROTO_PATH ?= "conmon-rs/common/proto"
-BINARY := conmon
+BINARY := conmonrs
```
saschagrunert, line 4: We have to change the artifact name there as well:
https://github.com/containers/conmon-rs/blob/0f113a1729538e24719968da5719e93d6d596393/.github/workflows/ci.yml#L51-L53

saschagrunert (commented):
I'm happy

saschagrunert (approved):

PR #290, opened 2022-04-04
Move version back to v0.1.0

saschagrunert:
Since the binary name changed, we now can go back to an unreleased version.

### DIFF
--- a/conmon-rs/server/src/version.rs
+++ b/conmon-rs/server/src/version.rs
@@ -1,6 +1,5 @@
 //! Generic version information for conmon
 
-use clap::crate_name;
 use getset::CopyGetters;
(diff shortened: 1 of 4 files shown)


conmon-rs/server/src/version.rs
```diff
@@ -40,7 +39,7 @@ impl Version {
 
     /// Print the version information to stdout.
     pub fn print(&self) {
-        println!("{} version {}", crate_name!(), build::PKG_VERSION);
+        println!("version: {}", build::PKG_VERSION);
```
saschagrunert, line 42: That was the non-breaking change to allow version parsing in c/common. I think we can skip that now.

rphillips (approved):

Issue #287, opened 2022-04-04
Implement container attach


saschagrunert:
To satisfy the CRI-O runtime method AttachContainer, a method should be created that requests an attach socket be created for a container or exec session.

PR #292, opened 2022-04-05
Add container attach skeleton

saschagrunert:
This adds the basic scaffolding for container attach as well as fixing
some test cases.

Refers to #287


saschagrunert:
This is the first iteration, PTAL @haircommander @rphillips

conmon-rs/server/src/attach.rs
```diff
@@ -0,0 +1,78 @@
...
+    path::Path,
+    thread,
+};
+
+#[derive(Debug)]
+/// Attach handles the attach socket IO of a container.
+pub struct Attach;
+
+impl Attach {
+    /// Createa a new attach instance.
```
haircommander, line 23: nit: `Create a`

conmon-rs/server/src/rpc.rs
```diff
@@ -61,6 +63,10 @@ impl conmon::Server for Server {
         let bundle_path = PathBuf::from(pry!(req.get_bundle_path()));
         let pidfile = bundle_path.join("pidfile");
         debug!("PID file is {}", pidfile.display());
+
+        let attach_path = bundle_path.join("attach");
+        self.attach = pry_err!(Attach::new(&attach_path).context("create attach endpoint")).into();
```
haircommander, line 68: we need an attach socket per container, can we store this in the child struct?

haircommander: (we'll eventually need an attach socket per Exec (not ExecSync) request as well)

conmon-rs/common/proto/conmon.capnp
```diff
@@ -24,6 +24,7 @@ interface Conmon {
 
     struct CreateContainerResponse {
         containerPid @0 :UInt32;
+        attachSocketPath @1 :Text;
```
haircommander, line 27: alternative to always returning as create container response: we create an AttachSocket request that a client could use to create an attach socket for the given create/exec session. I am open to either, what are your thoughts @saschagrunert

conmon-rs/server/src/attach.rs
saschagrunert: Fixed :+1:

conmon-rs/server/src/rpc.rs
saschagrunert: Fixed, moved into the reapable child.

conmon-rs/common/proto/conmon.capnp
saschagrunert: Yeah moved towards that :+1:

conmon-rs/common/proto/conmon.capnp
```diff
@@ -45,4 +45,16 @@ interface Conmon {
     }
 
     execSyncContainer @2 (request: ExecSyncContainerRequest) -> (response: ExecSyncContainerResponse);
+
+    ###############################################
+    # Attach
+    struct AttachContainerRequest {
```
haircommander, line 51: We could either add an optional ExecSession parameter here, or add a different endpoint. I am open to either. If we do the former, we may just want to call this "AttachRequest" as it'll be for container or exec

rphillips:
lgtm, after Peter's comment.

conmon-rs/common/proto/conmon.capnp
saschagrunert: I added a new `execSessionID` parameter to the request and renamed it to `AttachRequest/Response`

saschagrunert:
Ready for a new review round, PTAL @haircommander @rphillips

PR #298, opened 2022-04-05
exec: do not write to exit paths of container

haircommander:
as doing so will confuse clients who use the exit path as a signal that the container itself has exited


saschagrunert (approved):
LGTM

PR #303, opened 2022-04-06
WIP: Wire-in CRI logger

saschagrunert:
I'll stop with this PR now since it depends on the others.


conmon-rs/common/proto/conmon.capnp
```diff
@@ -20,6 +20,8 @@ interface Conmon {
         bundlePath @1 :Text;
         terminal @2 :Bool;
         exitPaths @3 :List(Text);
+        logPath @4 :Text;
+        maxLogSize @5 :Int64;
```
haircommander, line 24: to start, we can drop maxLogSize. Instead, we can implement a ReopenLog request that takes a container. I think in CRI-O we'll deprecate log size tracking outside of the kubelet (though we'll need it for podman)

saschagrunert:
The rebase takes too long, I'll come-up with a new PR.

PR #310, opened 2022-04-07
client, server: add support for very long server runtime dir paths

haircommander:
The runtime dir path will likely be the infra container's runDir, and thus something with a container UUID in it. Thus, it would be too long to hold the conmon.sock and work with the unix socket length. Work around this in the same way it was worked around in conmon (shout out to @giuseppe for coming up with the fix), where we use the `/proc/self/fd` entry of the parent directory of the target socket, and reference that when Dialing/Listening.

This also partially reverts https://github.com/containers/conmon-rs/pull/284, because for now runc doesn't support this functionality, so the process of receiving the console fd fails

### DIFF
--- a/conmon-rs/server/src/listener.rs
+++ b/conmon-rs/server/src/listener.rs
@@ -1,0 +1,32 @@
+use anyhow::{bail, Context, Result};
+use std::{
+    fs::File,
+    os::unix::io::AsRawFd,
+    path::{Path, PathBuf},
+};
+use tokio::net::UnixListener;
+
+pub fn bind_long_path(path: &Path) -> Result<UnixListener> {
+    let parent = match path.parent() {
+        Some(p) => p,
+        None => bail!(
+            "Tried to specify / as socket to bind to: {}",
+            path.display()
+        ),
+    };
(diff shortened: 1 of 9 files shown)


conmon-rs/server/src/listener.rs
```diff
@@ -0,0 +1,32 @@
...
+use tokio::net::UnixListener;
+
+pub fn bind_long_path(path: &Path) -> Result<UnixListener> {
+    let parent = match path.parent() {
+        Some(p) => p,
+        None => bail!(
+            "Tried to specify / as socket to bind to: {}",
+            path.display()
+        ),
+    };
```
saschagrunert, line 16: nit, you can use the Context on Option types too:
```suggestion
    let parent = path.parent().context(format!(
        "tried to specify / as socket to bind to: {}",
        path.display()
    ))?;
```

saschagrunert (approved):

conmon-rs/server/src/listener.rs
saschagrunert: See https://github.com/containers/conmon-rs/pull/312

PR #311, opened 2022-04-07
WIP: test: add stress test for ExecSyncRequests

haircommander:
Note: it sometimes takes a couple of runs to reproduce

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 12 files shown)


pkg/client/client_test.go
```diff
@@ -374,6 +375,51 @@ var _ = Describe("ConmonClient", func() {
 				Expect(result).NotTo(BeNil())
 				Expect(result.TimedOut).To(Equal(true))
 			})
+
+			testName = "should handle many requests"
+			if terminal {
+				testName += " with terminal"
+			}
+			FIt(testName, func() {
```
haircommander, line 383: WIP because of this line. I found it easier to debug with it

haircommander:
https://github.com/containers/conmon-rs/pull/300 also doesn't yet fix this problem :(

saschagrunert:
How does that work together with https://github.com/containers/conmon-rs/pull/300?

saschagrunert:
So, disconnecting the stdio streams in non terminal situations may be the problem here since it is globally applied. Does this work?

1. Crewte container with terminal
2. Exec sync without terminal
3. Exec sync with terminal

pkg/client/client_test.go
```diff
@@ -374,6 +375,51 @@ var _ = Describe("ConmonClient", func() {
...
+
+				// Wait for container to be running
+				Eventually(func() error {
+					return rr.RunCommandCheckOutput("running", "list")
+				}, time.Second*10).Should(BeNil())
+
+				var wg sync.WaitGroup
+				for i := 0; i < 15; i++ {
+					go func() {
+						wg.Add(1)
```
rphillips, line 409: the wg.Add should be prior to the go func since it'll race.

haircommander:
> Does it also break when `terminal == true`?
> 
> For non terminal scenarios, disconnecting the stdio streams may be the problem here since it is globally applied. We probably should do the setup only once.

terminal always succeeds! I'm not sure about disconnecting. we can do about 3 or 4 reliably, once it gets higher it usually stalls though

pkg/client/client_test.go
haircommander: ah good to know

saschagrunert:
We still use the single threaded runtime so it may be possible that we deadlock at some point:

https://github.com/containers/conmon-rs/blob/937f43846f297670feb741fa1998f579d70bcb49/conmon-rs/server/src/server.rs#L83

If changing the runtime to multi threaded solve the problem, then we would know that it’s indeed a thread dead lock.

haircommander:
> 

Good theory! Interestingly enough, we still hit this error if we set to multi_thread

rphillips:
Looks like we are deadlocking when terminal == false

haircommander:
> We are reusing a file descriptor... maybe we are not clearing a handle on it somewhere...

that would be for the terminal case, so I don't think it'd be the cause here. Unfortunately the non terminal conmonrs logs aren't printed to stdout because of the way it currently works. Also, FYI I believe that's both stdout and stderr streams being registered. The way the OCI runtime specifies terminal requires this to be the case

saschagrunert:
@haircommander the logs are still accessible via the journal

PR #313, opened 2022-04-08
Integrate container logging

saschagrunert:
Requires #300 

The commit e9dcc9fb5488801f953b88b3f92294388377c3ce is ready for review.

PTAL @haircommander @rphillips


conmon-rs/common/proto/conmon.capnp
```diff
@@ -20,6 +20,7 @@ interface Conmon {
         bundlePath @1 :Text;
         terminal @2 :Bool;
         exitPaths @3 :List(Text);
+        logPath @4 :Text;
```
haircommander, line 23: Eventually, we may need to support multiple log driver/log path pairs. currently, conmon just has them specified as `log_driver:/path`, but I think we could do better. maybe we could create a LogConfig and have a vector of those LogConfigs in the request?

haircommander:
I think this is a good start. We should make cri_logger optoinal, and have it be more generic (have a logger trait that cri logger satisfies). 

that, plus the protocol changes to allow for it to be generic

for now, let's merge and iterate (BTW: the logging tests we expect to pass pass when this is integrated into CRI-O)

haircommander:
@rphillips LGTY?

rphillips (approved):

conmon-rs/common/proto/conmon.capnp
saschagrunert: Good, following up on this in https://github.com/containers/conmon-rs/pull/315

PR #314, opened 2022-04-08
WIP: move streams implementation to Stdout instead of using conmonrs streams

haircommander:
plus a couple of other fixes attempting to fix the issues found in https://github.com/containers/conmon-rs/pull/311

note: we still don't pass the test found in https://github.com/containers/conmon-rs/pull/311, but I believe this is a step in the right direction

carrying https://github.com/containers/conmon-rs/pull/313


haircommander:
ok this is ready (though doesn't yet fix the test) and is carrying https://github.com/containers/conmon-rs/pull/320 so we probably could merge that too @rphillips @saschagrunert

PR #315, opened 2022-04-11
Add possibility to use multiple log drivers

saschagrunert:
We now add a configuration structure to allow multiple log drivers.

Follow-up on https://github.com/containers/conmon-rs/pull/313#discussion_r846278907


conmon-rs/server/src/rpc.rs
```diff
@@ -58,7 +58,11 @@ impl conmon::Server for Server {
         let id = pry!(req.get_id()).to_string();
         debug!("Got a create container request for id {}", id);
 
-        let log_path = pry!(req.get_log_path());
+        let log_path = pry!(pry_err!(pry!(req.get_log_drivers())
+            .iter()
+            .next()
+            .context("no log driver provided"))
+        .get_path());
```
saschagrunert, line 65: I expect that we create a higher level logging structure for parsing this which will be then used for writing and passing down the information to the lower level loggers.

saschagrunert:
Let's iterate, I continue on top of that. PTAL @rphillips

rphillips:
lgtm

PR #317, opened 2022-04-11
Use async read/write for attach

saschagrunert:
We now use async IO for attach as well as setting up the channels to read and write from them.

Refers to https://github.com/containers/conmon-rs/issues/287

### DIFF
--- a/conmon-rs/server/src/attach.rs
+++ b/conmon-rs/server/src/attach.rs
@@ -1,22 +1,42 @@
 use anyhow::{bail, Context, Result};
+use getset::Getters;
 use log::{debug, error};
 use nix::sys::socket::{
     bind, listen, socket, AddressFamily, SockAddr, SockFlag, SockType, UnixAddr,
 };
 use std::{
-    io::{BufReader, BufWriter, Read, Write},
     os::unix::{
         fs::PermissionsExt,
         io::{FromRawFd, RawFd},
-        net::{UnixListener, UnixStream},
+        net,
     },
     path::Path,
-    thread,
+};
+use tokio::{
+    io::{ErrorKind, Interest},
+    net::{UnixListener, UnixStream},
+    select,
+    sync::broadcast::{self, Receiver, Sender},
+    task,
 };
 
-#[derive(Clone, Debug)]
+/// The size of an attach packet.
+pub const ATTACH_PACKET_BUF_SIZE: usize = 8192;
+
+// TODO: remove this when being used
+#[allow(dead_code)]
+#[derive(Clone, Debug, Getters)]
 /// Attach handles the attach socket IO of a container.
-pub struct Attach;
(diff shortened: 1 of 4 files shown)


rphillips (approved):

conmon-rs/server/src/attach.rs
```diff
@@ -44,43 +49,82 @@ impl Attach {
...
+                    data.insert(0, Pipe::Stdout as u8);
+                    io.write_all(&data).await.context("write stdout")?;
+                },
+                res = stderr.recv() => {
+                    let mut data = res?;
+                    debug!("Got {} byte stderr data", data.len());
+                    data.insert(0, Pipe::Stderr as u8);
+                    io.write_all(&data).await.context("write stderr")?;
+                },
+                res = io.read_line(&mut line) => {
```
cgwalters, line 120: Careful with `select!` and stateful futures - https://blog.yoshuawuyts.com/futures-concurrency-3/

(In response to reading that blog I did https://github.com/ostreedev/ostree-rs-ext/pull/239 )

In this case the simplest fix may be to use e.g. https://docs.rs/tokio/latest/tokio/io/struct.Lines.html - then pin/box that future, and then you can read from that stream alongside the channels.

saschagrunert: Ugh, we should probably do that for the whole project at once. Created https://github.com/containers/conmon-rs/issues/318 for that as follow-up.

pkg/client/client_test.go
```diff
@@ -651,68 +651,8 @@ func testAttachSocketConnection(socketPath string) error {
 	}
 	defer conn.Close()
 
-	outputStream := io.WriteCloser(os.Stdout)
-	errorStream := io.WriteCloser(os.Stderr)
-
-	receiveStdout := make(chan error)
```
haircommander, line 657: why remove these pieces?

conmon-rs/server/src/attach.rs
cgwalters: Another option I think is to e.g. `Box::pin` the `io.read_line` future - that moves its state outside.

In general it's worth contrasting with e.g. Go - there every goroutine has a stack, which means one doesn't have to think about this type of problem.  But OTOH, cancellation is totally different and much more manual, and async Rust's "stackless" coroutine approach can be a lot more efficient in hot paths.

saschagrunert:
Had to rebase

rphillips:
[This](https://github.com/containers/conmon-rs/pull/328/commits/40374404b2f69eaf9d7e0316b226e9b15674b3a2#diff-c4d05538a581b262d00d4cafe443f75391046deb567f2acc1b59f1c56e14ca81R145-R150) tweak to the reader helps... trying it out over [here](https://github.com/containers/conmon-rs/pull/328/commits).

PR #319, opened 2022-04-11
Error if attach socket already exists

saschagrunert:
We should not silently remove the existing socket if it already exists.

Refers to https://github.com/containers/conmon-rs/issues/287


saschagrunert:
@rphillips PTAL

PR #320, opened 2022-04-11
Add container_log module

saschagrunert:
This makes the CRI logger optional as well as allowing to specify
multiple loggers.


saschagrunert:
@rphillips @haircommander PTAL :blush:

conmon-rs/server/src/container_log.rs
```diff
@@ -0,0 +1,84 @@
...
+            .collect();
+        Ok(Arc::new(RwLock::new(Self { drivers })))
+    }
+
+    /// Asynchronously initialize all loggers.
+    pub async fn init(&mut self) -> Result<()> {
+        for logger in self.drivers.iter_mut() {
+            match logger {
+                LogDriver::ContainerRuntimeInterface(ref mut cri_logger) => {
+                    cri_logger.init().await.context("init CRI logger")?
```
rphillips, line 50: Maybe collect the futures here so we can await them at the same time...

rphillips: might not be worth the effort

saschagrunert: Hm we would have to spawn a new task for them, right? I mean we could, but maybe we wanna re-consider this when actually having multiple log drivers :grin:

rphillips: let future = cri_logger.init();
futures.push(future);
tokio.join_all(futures).await

rphillips: Something like this... perhaps

rphillips:
```rust
diff --git a/conmon-rs/server/src/container_log.rs b/conmon-rs/server/src/container_log.rs
index 82cb83da..335f816f 100644
--- a/conmon-rs/server/src/container_log.rs
+++ b/conmon-rs/server/src/container_log.rs
@@ -56,13 +56,15 @@ impl ContainerLog {
 
     /// Reopen the container logs.
     pub async fn reopen(&mut self) -> Result<()> {
+        let mut futures = vec![];
         for logger in self.drivers.iter_mut() {
             match logger {
                 LogDriver::ContainerRuntimeInterface(ref mut cri_logger) => {
-                    cri_logger.reopen().await.context("reopen CRI logs")?
+                    futures.push(cri_logger.reopen());
                 }
             }
         }
+        tokio::join!(futures::future::join_all(futures));
         Ok(())
     }
 
```

haircommander: @rphillips fine with merging this and following up with that?

haircommander:
LGTM

PR #323, opened 2022-04-12
Join futures in container_log module

saschagrunert:
Follow-up on https://github.com/containers/conmon-rs/pull/320, where we
now join the futures instead of waiting for each one.


haircommander:
@saschagrunert can you rebase to get test fixes?

PR #325, opened 2022-04-12
Make integration tests work again

saschagrunert:
The test has to be excluded for now since the original issue has not been fixed yet.


saschagrunert:
Hm, it still does not seem to work :thinking:

PR #329, opened 2022-04-12
fix buffer read of zero

rphillips:
Signed-off-by: Ryan Phillips <rphillips@redhat.com>


haircommander:
there's still some issue with doing this (we still drop output sometimes), but I think it will end up being better than the stop broadcast stuff we have now (which I've concluded is really racy and is going to be hard to get right).

LGTM,but also I've opened an alternative that carries a variant of this [here](https://github.com/containers/conmon-rs/pull/330)

PR #330, opened 2022-04-12
streams/terminal: unify read loop

haircommander:
as they fundamentally shouldn't be very different.
    
Also, drop stop_rx from streams impl, as it is racy, include https://github.com/containers/conmon-rs/pull/329 and update a message from child_reaper

### DIFF
--- a/conmon-rs/server/src/streams.rs
+++ b/conmon-rs/server/src/streams.rs
@@ -1,17 +1,14 @@
 //! Pseudo terminal implementation.
 
 use crate::{
-    container_io::Message,
+    container_io::{ContainerIO, Message},
     container_log::{Pipe, SharedContainerLog},
 };
-use anyhow::{Context, Result};
+use anyhow::Result;
 use getset::{Getters, MutGetters};
-use log::{debug, error};
-use nix::errno::Errno;
-use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+use log::debug;
+use std::os::unix::io::AsRawFd;
 use tokio::{
-    fs::File,
-    io::{AsyncReadExt, BufReader},
     process::{ChildStderr, ChildStdout},
     sync::mpsc,
     task,
@@ -60,7 +57,7 @@ impl Streams {
 
         if let Some(stdout) = stdout {
             task::spawn(async move {
-                Self::read_loop(stdout.as_raw_fd(), Pipe::StdOut, logger, message_tx).await
+                ContainerIO::read_loop(stdout.as_raw_fd(), Pipe::StdOut, logger, message_tx).await
             });
         }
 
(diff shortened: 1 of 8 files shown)


haircommander:
also carry https://github.com/containers/conmon-rs/pull/327 for good measure

conmon-rs/server/src/streams.rs
```diff
@@ -42,120 +36,90 @@ impl Streams {
...
-                    Ok(n) if n > 0 => {
-                        debug!("Read {} bytes", n);
-                        let data = &buf[..n];
-
-                        let mut locked_logger = logger.write().await;
-                        locked_logger
-                            .write(pipe, data)
-                            .await
-                            .context("write to log file")?;
+            match reader.read(&mut buf).await {
```
haircommander, line 77: long term: we may want to switch to poll read? as it may include EAGAIN handling and make tokio happier

.github/workflows/ci.yml
```diff
@@ -171,4 +171,4 @@ jobs:
           key: ${{ runner.os }}-cargo-test-files-${{ hashFiles('pkg/client/files_test.go') }}
       - run: .github/install-deps
       - name: Integration tests
-        run: RUNTIME_PATH="/usr/sbin/runc" make integration-static
+        run: RUNTIME_PATH="/usr/sbin/runc" make integration
```
saschagrunert, line 174: Can we re-use the static binaries here?

pkg/client/client_test.go
```diff
@@ -257,7 +257,7 @@ var _ = Describe("ConmonClient", func() {
 	Describe("ExecSyncContainer", func() {
 		for _, terminal := range []bool{false} {
```
saschagrunert, line 258: Let's re-enable the terminal tests here.

conmon-rs/server/src/streams.rs
```diff
@@ -42,120 +36,90 @@ impl Streams {
...
+                    locked_logger
+                        .write(pipe, data)
+                        .await
+                        .context("write to log file")?;
+
+                    message_tx
+                        .send(Message::Data(data.into()))
+                        .context("send data message")?;
+                }
+                Ok(n) if n == 0 => {
```
saschagrunert, line 92: In the non-terminal case we always get one read with `n == 0` before the actual data comes in. We could work around this with something like:
```diff
diff --git a/conmon-rs/server/src/streams.rs b/conmon-rs/server/src/streams.rs
index 9e3c28a..1cdb10a 100644
--- a/conmon-rs/server/src/streams.rs
+++ b/conmon-rs/server/src/streams.rs
@@ -73,6 +73,7 @@ impl Streams {
         let stream = unsafe { File::from_raw_fd(fd) };
         let mut reader = BufReader::new(stream);
         let mut buf = vec![0; 1024];
+        let mut already_read = false;
         loop {
             match reader.read(&mut buf).await {
                 Ok(n) if n > 0 => {
@@ -88,8 +89,10 @@ impl Streams {
                     message_tx
                         .send(Message::Data(data.into()))
                         .context("send data message")?;
+
+                    already_read = true;
                 }
-                Ok(n) if n == 0 => {
+                Ok(n) if n == 0 && already_read => {
                     debug!("No more to read");
 
                     message_tx
```

haircommander: do you know why that is?

pkg/client/client_test.go
haircommander: oops, good catch, I did so here and https://github.com/containers/conmon-rs/pull/327

conmon-rs/server/src/streams.rs
saschagrunert: I have no idea unfortunately.

haircommander: do you think it was what @rphillips found where we sent Message::Done twice?

haircommander:
I would say overall this leaves things in a better shape than they were. I vote to merge and iterate (unless there are concerns :smiley: )

pkg/client/client_test.go
```diff
@@ -294,15 +294,17 @@ var _ = Describe("ConmonClient", func() {
...
 						result, err := sut.ExecSyncContainer(context.Background(), &client.ExecSyncConfig{
 							ID:       ctrID,
-							Command:  []string{"/busybox", "echo", "-n", "hello", "world"},
+							Command:  []string{"/busybox", "echo", "-n", "hello", "world", fmt.Sprintf("%d", i)},
 							Terminal: terminal,
 							Timeout:  timeoutUnlimited,
 						})
 						Expect(err).To(BeNil())
 						Expect(result).NotTo(BeNil())
+						Expect(result).To(Equal(fmt.Sprintf("hello world %d", i)))
```
rphillips, line 307: My patch, but this should be result.Stdout and line 258 should be Describe

PR #332, opened 2022-04-13
Integrate attach

saschagrunert:
Integrate the attach functionality.

Fixes https://github.com/containers/conmon-rs/issues/287

### DIFF
--- a/conmon-rs/server/src/attach.rs
+++ b/conmon-rs/server/src/attach.rs
@@ -1,25 +1,88 @@
+use crate::container_io::Pipe;
 use anyhow::{bail, Context, Result};
 use log::{debug, error};
 use nix::sys::socket::{
     bind, listen, socket, AddressFamily, SockAddr, SockFlag, SockType, UnixAddr,
 };
 use std::{
-    io::{BufReader, BufWriter, Read, Write},
     os::unix::{
         fs::PermissionsExt,
         io::{FromRawFd, RawFd},
-        net::{UnixListener, UnixStream},
+        net,
     },
-    path::Path,
-    thread,
+    path::{Path, PathBuf},
+    sync::Arc,
 };
+use tokio::{
+    io::{ErrorKind, Interest, Ready},
+    net::{UnixListener, UnixStream},
+    sync::RwLock,
+    task,
+    time::{timeout, Duration},
+};
+
+#[derive(Debug, Clone, Default)]
+/// A shared container attach abstraction.
+pub struct SharedContainerAttach(Arc<RwLock<Vec<Attach>>>);
+
+impl SharedContainerAttach {
+    /// Add a new attach endpoint to this shared container attach instance.
+    pub async fn add(&self, attach: Attach) {
+        self.0.write().await.push(attach);
+    }
+
+    /// Try to read from all attach endpoints standard input and return the first result.
+    pub async fn try_read(&self) -> Result<Option<Vec<u8>>> {
+        self.cleanup().await;
+        for attach in self.0.read().await.iter() {
+            if let Some(data) = attach.try_read().await? {
+                return Ok(data.into());
+            }
+        }
+        Ok(None)
+    }
+
+    /// Write a buffer to all attach endpoints.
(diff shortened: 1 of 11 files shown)


conmon-rs/server/src/attach.rs
```diff
@@ -46,43 +59,134 @@ impl Attach {
...
+    where
+        T: AsRef<[u8]>,
+    {
+        let packets = buf
+            .as_ref()
+            .chunks(ATTACH_PACKET_BUF_SIZE - 1)
+            .map(|x| {
+                let mut y = x.to_vec();
+                let p = match pipe {
+                    Pipe::StdOut => 2,
```
haircommander, line 166: we should have a TODO here to not go over stdout/stderr but the container's pipes

haircommander: given how this works, should we save the container's fds in the container_io instance and save that in the child instance?

saschagrunert: I think in long term we have to store the container IOs (the one from createcontainer and all execs) in the child and reference them by IDs. This way we can store the attach inside of the corresponding container IO on the attach RPC.

The `try_read` and `write` functions will be then called by the global read loop.

saschagrunert:
This is ready for review independently of the failing integration tests. I think testing is bit hard here. We would have to create a container producing actual output to attach 🤔

conmon-rs/server/src/attach.rs
```diff
@@ -46,43 +102,125 @@ impl Attach {
...
-    fn handle_client(stream: UnixStream) -> Result<()> {
-        debug!("Got new attach stream connection");
+    /// Try to read from all streams standard input and return the first result.
+    pub async fn try_read(&self) -> Result<Option<Vec<u8>>> {
+        let mut cleanup_idxs = vec![];
+        let mut clients = self.clients.write().await;
 
-        let mut reader = BufReader::new(&stream);
-        let mut writer = BufWriter::new(&stream);
+        for (idx, stream) in clients.iter().enumerate() {
```
haircommander, line 135: this will mean if we've registered multiple clients, the first will get priority for stdin, right? FYI @mheon for podman integration

conmon-rs/server/src/container_io.rs
```diff
@@ -104,11 +110,24 @@ impl ContainerIO {
...
         logger: SharedContainerLog,
         message_tx: UnboundedSender<Message>,
+        attach: SharedContainerAttach,
     ) -> Result<()> {
         let stream = unsafe { File::from_raw_fd(fd) };
+        let mut writer = unsafe { File::from_raw_fd(fd) };
         let mut reader = BufReader::new(stream);
         let mut buf = vec![0; 1024];
         loop {
+            if let Some(data) = attach
```
haircommander, line 120: we should register a new routine for stdin processing when we register the read loops IMO

pkg/client/client_test.go
```diff
@@ -741,68 +741,49 @@ func testAttachSocketConnection(socketPath string) error {
...
-				fmt.Fprintln(os.Stderr, err)
-				break
-			}
-		}
+	return nil
 
-		fmt.Printf("Read %v bytes\n", n)
-		if n > 0 {
-			var dst io.Writer
+	/*
```
haircommander, line 765: eventually when we figure out this code, I think it should live in client. Ideally, we could provide an attach that the caller gives channels (And some other args, probably in a struct) and the client wires everything up

conmon-rs/server/src/attach.rs
saschagrunert: Yes that's correct.

conmon-rs/server/src/container_io.rs
saschagrunert: Absolutely, I'll create a dedicated task for that :+1:

pkg/client/client_test.go
saschagrunert: Yeah we can follow-up on that as well :+1:

saschagrunert:
@rphillips @haircommander PTAL again

haircommander:
LGTM

haircommander:
oops needs rebase now

PR #338, opened 2022-04-14
 client: restore server if already running 

haircommander:
by returning server pid in version command

plus carrying https://github.com/containers/conmon-rs/pull/337 and updating gitignore


pkg/client/client.go
```diff
@@ -222,6 +227,29 @@ func (c *ConmonClient) Version(ctx context.Context) (string, error) {
 	return response.Version()
 }
 
+func (c *ConmonClient) processIDFromServer(ctx context.Context) (uint32, error) {
```
rphillips, line 230: This function should be exported. If the proto is in the public API we should have a public client for it.

haircommander: I see your point, but currently Version() endpoint doesn't actually return all of the version information, only the version string. I am of the opinion that we should either return all of that information in Version() (which can be done as a follow up) or go with this approach where we keep some functionality internal to the client because it's an implementation detail between client and server

rphillips: ah, I didn't realize Version() is only returning the string. Let's do the followup

rphillips (approved):

saschagrunert (approved):

PR #346, opened 2022-04-21
Cleanup some types and code

saschagrunert:
We now cleanup some type conversions as well as method signatures.


saschagrunert:

PR #347, opened 2022-04-21
Find default runtime in $PATH

saschagrunert:
We now fallback to a default runtime and try to find them in $PATH if
nothing else is provided.

### DIFF
--- a/conmon-rs/server/src/config.rs
+++ b/conmon-rs/server/src/config.rs
@@ -1,10 +1,13 @@
 //! Configuration related structures
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use clap::{AppSettings, Parser};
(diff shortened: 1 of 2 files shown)


rphillips (approved):

conmon-rs/server/src/config.rs
```diff
@@ -124,9 +127,19 @@ impl Default for Config {
 const SOCKET: &str = "conmon.sock";
 const PIDFILE: &str = "pidfile";
 
+const DEFAULT_RUNTIME: &str = "runc";
```
haircommander, line 130: I'm unsure about this. as we move to support podman, their default is crun. we don't have a config value to draw this from. I think we may be better not providing a default

saschagrunert: Okay sounds good to me.

PR #352, opened 2022-04-26
client,server: formalize attach behavior

haircommander:
podman and cri-o currently have this code spread across the two repositories. This causes code drift and duplication. We should really move the code that uses the attach closer to the entity that creates the socket. Then, conmon-rs client and server can be in control of the protocol.

also fix some issues that come from a long bind path for the attach socket


pkg/client/attach.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+		stdinDone <- err
+	}()
+
+	return receiveStdoutError, stdinDone
+}
+
+func redirectResponseToOutputStreams(cfg *AttachConfig, conn io.Reader) error {
+	var err error
+	buf := make([]byte, 8192+1) /* Sync with conmon STDIO_BUF_SIZE */
```
saschagrunert, line 144: nit: `8192` should be a const on top of the file

pkg/client/attach.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+type TerminalSize struct {
+	Width, Height uint16
+}
+
+type AttachStreams struct {
+	Stdin                                   io.Reader
+	Stdout, Stderr                          io.WriteCloser
+	AttachStdin, AttachStdout, AttachStderr bool
```
saschagrunert, line 29: Could we make those implicit by checking `Stdin/Stdout/Stderr == nil` so we can remove them?

pkg/client/attach.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/containers/podman/v3/pkg/kubeutils"
+	"github.com/containers/podman/v3/utils"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	AttachPipeStdin  = 1
+	AttachPipeStdout = 2
+	AttachPipeStderr = 3
```
saschagrunert, line 19: I think we do not have to export them.

pkg/client/attach.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+type AttachConfig struct {
+	ID, SocketPath, ExecSession   string
+	Tty, DetachStdin, Passthrough bool
+	Resize                        chan define.TerminalSize
+	AttachReady                   chan bool
+	Streams                       AttachStreams
+	StartFunc                     func() error
+	DetachKeys                    []byte
```
saschagrunert, line 39: I'd love to see doc comments on those values.

pkg/client/attach.go
```diff
@@ -0,0 +1,257 @@
...
+			return <-receiveStdoutError
+		}
+	}
+	return nil
+}
+
+type SetWindowSizeContainerConfig struct {
+	ID     string
+	Width  uint16
+	Height uint16
```
saschagrunert, line 222: Let's add comments here too, it's the height/width in rows/columns.

saschagrunert (commented):
Just a bunch of nits, otherwise LGTM

pkg/client/attach.go
```diff
@@ -0,0 +1,257 @@
...
+	return c.attach(ctx, cfg)
+}
+
+func (c *ConmonClient) attach(ctx context.Context, cfg *AttachConfig) error {
+	var (
+		conn *net.UnixConn
+		err  error
+	)
+	if !cfg.Passthrough {
+		logrus.Debugf("Attaching to container %s", cfg.ID)
```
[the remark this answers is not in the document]
haircommander: good idea!

pkg/client/attach.go
saschagrunert: How about adding a new structure for them which basically wraps the interface internally?

pkg/client/attach.go
haircommander: I just dropped the struct

pkg/client/attach.go
haircommander: done!

pkg/client/attach.go
haircommander: what did you have in mind?

haircommander:
most nits addressed @saschagrunert

pkg/client/attach.go
saschagrunert: > what did you have in mind?

I'll follow-up on that. Let's discuss it there.

saschagrunert (approved):

PR #358, opened 2022-04-28
Wrap stdio in structs for attach

saschagrunert:
This allows that the values can be nil and makes the `Attach*` booleans obsolete.

### DIFF
--- a/pkg/client/attach.go
+++ b/pkg/client/attach.go
@@ -22,9 +22,22 @@ const (
 )
 
 type AttachStreams struct {
-	Stdin                                   io.Reader
-	Stdout, Stderr                          io.WriteCloser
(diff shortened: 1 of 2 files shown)


haircommander:
LGTM, good idea

pkg/client/attach.go
```diff
@@ -22,9 +22,26 @@ const (
...
+	io.Reader
+}
+
+type Stdout struct {
+	io.WriteCloser
+}
+
+type Stderr struct {
+	io.WriteCloser
 }
```
haircommander, line 45: actually, nit, do we reallly need two structs here? we could just name it in and out and then use  the correct ones?

haircommander:
thank you!

PR #360, opened 2022-04-28
Move `rpc` and `server` into same module

saschagrunert:
This allows making more server methods private as well as exposing only necessary APIs.

We also now use `.clone()` instead of `Arc::clone`.


conmon-rs/server/src/child_reaper.rs
```diff
@@ -46,7 +47,7 @@ macro_rules! lock {
 
 impl ChildReaper {
     pub fn get(&self, id: &str) -> Result<ReapableChild> {
-        let locked_grandchildren = Arc::clone(&self.grandchildren);
+        let locked_grandchildren = &self.grandchildren().clone();
```
haircommander, line 50: there was a reason we went for Arc::clone, @cgwalters I believe it was yours

rphillips: both syntaxes are correct. [docs](https://doc.rust-lang.org/std/sync/struct.Arc.html#deref-behavior).

I do prefer the .clone() one. It looks more idiomatic to me.

haircommander:
I'm flip flopping on this one. On  the one hand, it's good to have a cleaner public interface. on the other hand, I think it makes the server.rs file hard to parse. If we could break it itno multiple files and have them be in the same module, I think that'd be better, but since rust doesn't allow it, I am leaning more against this. I am open to merging it if y'all feel strongly for it though

rphillips:
Agreed. I'm on the fence too. I sorta like the rpc split out. Perhaps we put the rpc.rs file back and keep the clone changes?

saschagrunert:
Found a way to make fields crate public, even with using getset. Now we can move everything back into the `rpc` crate while still cleaning the API surface. PTAL again @haircommander @rphillips

haircommander:
neat!

haircommander:
thanks!

PR #364, opened 2022-04-28
Make client logging configurable

saschagrunert:
A new configuration option allows to use a different logger rather than
the `logrus.StandardLogger()`.


haircommander:
good idea!

PR #382, opened 2022-05-03
Wrap errors and enable more linters

saschagrunert:
We now enable the linters cyclop, gomnd, gosec and wrapcheck and fix their corresponding lints.


haircommander:
LGTM

PR #386, opened 2022-05-04
Add docs to reopen container logs RPC

saschagrunert:
The docs were missing from my previous patch, so we now ensure that this method has a docstring as well.


saschagrunert:

PR #406, opened 2022-05-10
Replace `.unwrap`s

saschagrunert:
We now replace the unwrap code parts with case by case handling.


haircommander:
LGTM

rphillips:
Thanks for fixing these up.

PR #409, opened 2022-05-10
Make child reaper private again

saschagrunert:
We do not have to expose it via the public API so we make it crate public again via getset.


rphillips:
woops!

PR #413, opened 2022-05-12
Sanitize oom watcher logs

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now apply a bunch of rules, for example:

- Start logs upper case
- Remove redundant information like the PID (already part of the tracing span)
- Do not print "Error" for error logs, since the level already tells us
  that it's an error.

None
None

```release-note
None
```


haircommander:
LGTM

saschagrunert:
@haircommander it looks like that tide is not active yet for that project :thinking:

PR #418, opened 2022-05-17
Make server config private

saschagrunert:
#### What type of PR is this?

/kind cleanup

The returned type is private so the method is unusable anyhow. Since we
cannot set the configuration from the outside we now make it private to
lower the API surface.

None
None

```release-note
None
```


rphillips:
/lgtm

Let's see if Tide merges.

PR #466, opened 2022-06-14
not wait for stdout complete

### LINKED ISSUE #410: critest attach test fails
The critest conformance test `runtime should support attach [Conformance]` fails by timing out:

https://github.com/kubernetes-sigs/cri-tools/blob/1a0b272405a694d7d82606c5a7eddd7905321c4b/pkg/validate/streaming.go#L109-L122

Right now the test got excluded by CRI-O: https://github.com/cri-o/cri-o/blob/db76322414a7888c7642b707e07dbb9905aa969f/.github/workflows/integration.yml#L57

We either have to fix that on the CRI-O or the conmon-rs side. From the logs I see that conmonrs gets the data from the test:

```
May 11 10:41:06 nixos conmonrs[504862]: Got a attach container request
May 11 10:41:06 nixos conmonrs[504862]: Creating attach socket: /var/lib/containers/storage/overlay-containers/c16e58ee6177513959550004b949d71472b0cca52ca603bd3764195d22590b59/userdata/ee14eb08f450ffda35dd689411663b8b19d3d7947e695844e49ecca5ce32dc0d/attach
May 11 10:41:06 nixos conmonrs[504862]: Start listening on attach socket
May 11 10:41:06 nixos conmonrs[504862]: Got new attach stream connection
May 11 10:41:07 nixos conmonrs[504862]: Read 14 stdin bytes from client
May 11 10:41:07 nixos conmonrs[504862]: fd:23:read 5 bytes
May 11 10:41:08 nixos conmonrs[504862]: Wrote stdout packet to client
```

kangclzjc:
Signed-off-by: Kang.Zhang <Kang.zhang@intel.com>

#### What type of PR is this?
/kind bug

critest hung at attach, currently crio skip this test
Fixes #410 

None

```release-note
None
```

### DIFF
--- a/pkg/client/attach.go
+++ b/pkg/client/attach.go
@@ -155,11 +155,6 @@ func (c *ConmonClient) attach(ctx context.Context, cfg *AttachConfig) (err error
 		if err != nil {
 			return fmt.Errorf("failed to connect to container's attach socket: %v: %w", cfg.SocketPath, err)
 		}
-		defer func() {
-			if err := conn.Close(); err != nil {
-				c.logger.Errorf("unable to close socket: %q", err)
-			}
-		}()
 	}
 
 	if cfg.PreAttachFunc != nil {
@@ -191,6 +186,11 @@ func (c *ConmonClient) setupStdioChannels(
 ) (receiveStdoutError, stdinDone chan error) {
 	receiveStdoutError = make(chan error)
 	go func() {
+		defer func() {
+			if err := conn.Close(); err != nil {
+				c.logger.Errorf("unable to close socket: %q", err)
+			}
+		}()
 		receiveStdoutError <- c.redirectResponseToOutputStreams(cfg, conn)
 	}()
 
@@ -295,9 +295,6 @@ func (c *ConmonClient) readStdio(
 			if connErr := conn.CloseWrite(); connErr != nil {
 				c.logger.Errorf("Unable to close conn: %v", connErr)
 			}
-		}
-		if cfg.Streams.Stdout != nil || cfg.Streams.Stderr != nil {
-			return <-receiveStdoutError
 		}
 	}
 


pkg/client/attach.go
```diff
@@ -296,9 +296,6 @@ func (c *ConmonClient) readStdio(
 				c.logger.Errorf("Unable to close conn: %v", connErr)
 			}
 		}
-		if cfg.Streams.Stdout != nil || cfg.Streams.Stderr != nil {
-			return <-receiveStdoutError
-		}
```
haircommander, line 301: hm I don't see the reasoning for dropping this

haircommander:
thanks @kangclzjc !

pkg/client/attach.go
kangclzjc: In conmonrs, receiveStdoutError will never have value.  In conmonrs, even conn.Read get output result, it will never get an EOF error, so this loop will be running forever. But in conmon, after read output 'hello', it will get an EOF error, and break this loop. It seems even we finish exec command from stdin, we still need to try to read from conn ?
```
func (c *ConmonClient) redirectResponseToOutputStreams(cfg *AttachConfig, conn io.Reader) (err error) {
	buf := make([]byte, attachPacketBufSize+1) /* Sync with conmonrs ATTACH_PACKET_BUF_SIZE */
	for {
		nr, er := conn.Read(buf)
```

haircommander: even if we're absorbing the EOF [here](https://github.com/containers/conmon-rs/blob/main/pkg/client/attach.go#L244) we still want to wait to process all of the output, even if there's no input left. For instance, if the command is

`cat` and the stdin is `/dev/random\n` we'd want to see the output of /dev/random

PR #508, opened 2022-07-06
Switch to c/common from podman/v4

### LINKED ISSUE #504: go.mod: do not depend on github.com/containers/podman
Given that podman likely wants to import this go client lib as well (https://github.com/containers/podman/pull/14804) this lib should not depend on any podman code. While it can work the circular dependency is just asking for trouble when podman tries to make breaking changes to the code.

I propose that we move the required podman code to c/common and then podman and conmon-rs can use it without trouble.

saschagrunert:
#### What type of PR is this?

/kind cleanup
/kind dependency-change

The source code has now being moved over to c/common which we now rely on.

Fixes #504 
None

```release-note
None
```


Luap99 (commented):
LGTM, I was just about to make the same change.

PR #540, opened 2022-07-21
Add support for max log size

saschagrunert:
#### What type of PR is this?

/kind feature

The CRI logger already supports the max log size, which is now exposed via the capnproto API and the client.

None
None

```release-note
Added support for max log size.
```


conmon-rs/common/proto/conmon.capnp
```diff
@@ -26,9 +26,15 @@ interface Conmon {
...
 
     struct LogDriver {
+        # The type of the log driver.
         type @0 :Type;
+
+        # The filesystem path of the log driver, if required.
         path @1 :Text;
 
+        # The maximum log size in bytes, 0 means unlimited.
+        maxSize @2 :UInt32;
```
rphillips, line 36: Let's future proof this and make the maxSize a :UInt64.

pkg/client/client.go
```diff
@@ -438,6 +438,10 @@ type LogDriver struct {
 
 	// Path specifies the filesystem path of the log driver.
 	Path string
+
+	// MaxSize is the maximum amount of bytes to be written before rotation.
+	// 0 translates to an unlimited size.
+	MaxSize uint32
```
rphillips, line 444: ditto

conmon-rs/common/proto/conmon.capnp
saschagrunert: Yep, done :+1:

pkg/client/client.go
saschagrunert: Done

PR #543, opened 2022-07-25
Fix IO busy loops

saschagrunert:
#### What type of PR is this?

/kind bug

Both loops do not block the read, especially when no attach session is happening right now or the container produces no output. We fix that behavior by polling every 100ms.

None
None

```release-note
Fixed busy loops for reading container IO.
```


saschagrunert:

haircommander:
/lgtm

maybe long term we can consider using epoll or something (though we may end up busy looping on that :upside_down_face: )

PR #552, opened 2022-07-25
Add support for `--cgroup-manager`

saschagrunert:
#### What type of PR is this?

/kind bug

We now can select a cgroup manager and set `--systemd-cgroup` on the OCI runtime level if systemd is selected (the default).

This fixes the passing of systemd related annotations to the config, for example: https://github.com/opencontainers/runc/pull/2224

None
None

```release-note
Using `--systemd-cgroup` for runtime commands.
```

### DIFF
--- a/conmon-rs/server/src/config.rs
+++ b/conmon-rs/server/src/config.rs
@@ -88,6 +88,18 @@ pub struct Config {
     )]
     /// Do not fork if true
     skip_fork: bool,
+
+    #[get_copy = "pub"]
+    #[clap(
+        default_value(CgroupManager::Systemd.into()),
+        env(concat!(prefix!(), "CGROUP_MANAGER")),
+        long("cgroup-manager"),
+        short('c'),
+        possible_values(CgroupManager::iter().map(|x| x.into()).collect::<Vec<&str>>()),
+        value_name("MANAGER")
+    )]
+    /// Select the cgroup manager to be used
+    cgroup_manager: CgroupManager,
 }
 
 #[derive(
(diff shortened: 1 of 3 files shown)


saschagrunert:
PTAL @rphillips @haircommander

conmon-rs/server/src/server.rs
```diff
@@ -222,6 +222,8 @@ impl Server {
         }
     }
 
+    const SYSTEMD_CGROUP_ARG: &'static str = "--systemd-cgroup";
```
haircommander, line 225: we I think should actually toggle this based on the passed cgroup manager I think. Maybe a flag at startup, as we'll use one cgroup manager per pod (more acurately per node for cri-o, but possibly per pod for podman)

saschagrunert: Yep, added it via a configuration variable.

pkg/client/client.go
```diff
@@ -225,6 +228,10 @@ func (c *ConmonClient) toArgs(config *ConmonServerConfig) (entrypoint string, ar
 		args = append(args, "--log-driver", config.LogDriver)
 	}
 
+	if config.SystemdCgroup {
+		args = append(args, "--systemd-cgroup")
```
rphillips, line 232: nit: let's make this a const as well.

saschagrunert: Fixed as suggested :+1:

rphillips:
lgtm

I'll defer to @haircommander for the final lgtm

conmon-rs/server/src/config.rs
```diff
@@ -88,6 +88,15 @@ pub struct Config {
     )]
     /// Do not fork if true
     skip_fork: bool,
+
+    #[get_copy = "pub"]
+    #[clap(
+        env(concat!(prefix!(), "SYSTEMD_CGROUP")),
+        long("systemd-cgroup"),
+        value_name("SYSTEMD_CGROUP")
```
haircommander, line 96: Thinking about this a bit more, do we want to use `cgroup-manager` here? it could give us more flexibility in the future, but also may be unnecessary

haircommander, line 96: Thinking about this a bit more, do we want to use `cgroup-manager` here? it could give us more flexibility in the future, but also may be unnecessary

saschagrunert: Yep, done :+1:

saschagrunert: Yep, done :+1:

rphillips:
Peter is AFK. Going to lgtm and we can iterate if we would like to change this PR further.

/lgtm

rphillips:
miss click

/lgtm

PR #562, opened 2022-07-27
Fix CRI logger max size behavior

saschagrunert:
#### What type of PR is this?

/kind bug

We should account the bytes written globally and not on a per `write` basis. Added a unit test for the erroneous behavior.

None
None

```release-note
Fixed CRI logger max size behavior.
```


saschagrunert:

Issue #620, opened 2022-08-11
Attach detaches from container too early


jakecorrenti:
### What happened?

I am currently trying to integrate conmon-rs into Podman. I am trying to get the attach functionality working. When I run the command `sudo bin/podman run --log-driver=k8s-file --name=demo -d alpine top`, the container successfully starts and the logs display that there are bytes being read: 
```
2022-08-11T18:14:42.317436Z DEBUG backend:create_container{container_id="a186f94a8b5cd7953e3fe191a0bbf2dfbb8e4199c2dd3c1dd3a05c6ac44674f5" uuid="1c4c62f4-c230-485f-895e-aee634fe43a6"}:promise:stdout: conmonrs::container_io: 238: Read 304 bytes
2022-08-11T18:14:47.322829Z DEBUG backend:create_container{container_id="a186f94a8b5cd7953e3fe191a0bbf2dfbb8e4199c2dd3c1dd3a05c6ac44674f5" uuid="1c4c62f4-c230-485f-895e-aee634fe43a6"}:promise:stdout: conmonrs::container_io: 238: Read 304 bytes
2022-08-11T18:14:52.327820Z DEBUG backend:create_container{container_id="a186f94a8b5cd7953e3fe191a0bbf2dfbb8e4199c2dd3c1dd3a05c6ac44674f5" uuid="1c4c62f4-c230-485f-895e-aee634fe43a6"}:promise:stdout: conmonrs::container_io: 238: Read 304 bytes
```
When I try and attach to the container doing `sudo bin/podman attach demo`, it attaches and briefly displays the output of top: 
```
Mem: 15176388K used, 50390388K free, 1378696K shrd, 4192K buff, 9559136K cached
CPU:   1% usr   0% sys   0% nic  98% idle   0% io   0% irq   0% sirq
Load average: 0.26 0.45 0.47 1/1736 1
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
    1     0 root     R     1604   0%  10   0% top
```
However, after this output displays, the container immediately detaches even though the logs show that conmon-rs is still reading bytes.

Log output when attaching: 
```
2022-08-11T18:16:01.775594Z DEBUG backend: conmonrs::rpc: 50: Got a version request
[...]

PR #683, opened 2022-09-02
cancel I/O (notably attach) when container exits

haircommander:
#### What type of PR is this?

/kind bug
Previously, the attach code was sending Done too early. This caused clients to detach early while there was still output to process. It previously used the "no output" condition to decide that there was no more output. However, it seems the read pipe will send `n == 0` (which I think should be looked into separately)

Really, the condition to stop the attach are two-fold: client detaches, or container exits. the former case should be handled. This PR adds the latter case by using a cancellation token and wiring it throughout the I/O code.

A followup should be checking that we're not CPU cycling when we get no output from the pipes (`n == 0` cases that were dropped) 
fixes https://github.com/containers/conmon-rs/issues/620

```release-note
Fix a bug where the server would detach from an attach session too early
```

### DIFF
--- a/conmon-rs/server/src/attach.rs
+++ b/conmon-rs/server/src/attach.rs
@@ -257,7 +257,7 @@ impl Attach {
             select! {
                 res = rx.recv() => {
                     let (pipe, buf) = res?;
-                    let mut packets = buf
+                    let packets = buf
                         .chunks(Self::PACKET_BUF_SIZE - 1)
                         .map(|x| {
                             let mut y = x.to_vec();
@@ -270,7 +270,6 @@ impl Attach {
                             y
                         })
                         .collect::<Vec<_>>();
-                    packets.push(Self::DONE_PACKET.to_vec());
 
                     let len = packets.len() - 1;
                     for (idx, packet) in packets.iter().enumerate() {
@@ -286,6 +285,13 @@ impl Attach {
                 }
                 _ = token.cancelled() => {
                     debug!("Exiting because token cancelled");
+                    match write_half.write(Self::DONE_PACKET).await {
+                        Ok(_) => {
+                            debug!("Wrote done packet to client")
+                        }
+                        Err(ref e) if e.kind() == ErrorKind::WouldBlock || e.kind() == ErrorKind::BrokenPipe => {},
+                        Err(e) => bail!("unable to write done packet: {:#}", e),
+                    }
                     return Ok(());
                 }
             }
--- a/conmon-rs/server/src/attach.rs
+++ b/conmon-rs/server/src/attach.rs
@@ -22,9 +22,11 @@ use tokio::{
         unix::{OwnedReadHalf, OwnedWriteHalf},
         UnixListener,
     },
+    select,
     sync::broadcast::{self, Receiver, Sender},
     task,
 };
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, debug_span, error, Instrument};
 
 #[derive(Debug)]
@@ -59,7 +61,7 @@ impl Clone for SharedContainerAttach {
 
 impl SharedContainerAttach {
     /// Add a new attach endpoint to this shared container attach instance.
-    pub async fn add<T>(&mut self, socket_path: T) -> Result<()>
+    pub async fn add<T>(&mut self, socket_path: T, token: CancellationToken) -> Result<()>
     where
         T: AsRef<Path>,
         PathBuf: From<T>,
@@ -68,6 +70,7 @@ impl SharedContainerAttach {
             socket_path,
             self.read_half_tx.clone(),
             self.write_half_tx.clone(),
+            token,
         )
         .context("create attach endpoint")
     }
@@ -110,6 +113,7 @@ impl Attach {
         socket_path: T,
         read_half_tx: Sender<Vec<u8>>,
         write_half_tx: Sender<(Pipe, Vec<u8>)>,
+        token: CancellationToken,
     ) -> Result<()>
     where
         T: AsRef<Path>,
@@ -144,7 +148,7 @@ impl Attach {
 
         task::spawn(
             async move {
-                if let Err(e) = Self::start(fd, read_half_tx, write_half_tx).await {
+                if let Err(e) = Self::start(fd, read_half_tx, write_half_tx, token).await {
                     error!("Attach failure: {:#}", e);
                 }
             }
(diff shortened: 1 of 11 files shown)


conmon-rs/server/src/attach.rs
```diff
@@ -188,74 +195,92 @@ impl Attach {
...
 
-    async fn read_loop(mut read_half: OwnedReadHalf, tx: Sender<Vec<u8>>) -> Result<()> {
+    async fn read_loop(mut read_half: OwnedReadHalf, tx: Sender<Vec<u8>>, token: CancellationToken) -> Result<()> {
         loop {
             let mut buf = vec![0; Self::PACKET_BUF_SIZE];
-            match read_half.read(&mut buf).await {
-                Ok(n) if n > 0 => {
-                    if let Some(first_zero_idx) = buf.iter().position(|&x| x == 0) {
-                        buf.resize(first_zero_idx, 0);
+            select! {
```
haircommander, line 201: summoning @cgwalters

I know you linked something in the past about how neferious select statements can be. I can't frankly see a better way to handle this. If you have any tips or a way to absolve the risk of select, I would appreciate it :)

conmon-rs/server/src/container_io.rs
```diff
@@ -162,9 +176,15 @@ impl ContainerIO {
...
-                    Self::read_stream_with_timeout(time_to_timeout, t.message_rx_mut()).await;
-                (stdout, vec![], timed_out)
+                if let Some(message_rx) = t.message_rx_mut() {
+                    let (stdout, timed_out) =
+                        Self::read_stream_with_timeout(time_to_timeout, message_rx).await;
+                    (stdout, vec![], timed_out)
+                } else {
+                    // TODO FIXME
+                    error!("read_all_with_timeout called before message_rx was registered");
+                    (vec![], vec![], false)
```
saschagrunert, line 186: I'd say we `bail!` here and return a `Result<(Vec<u8>, Vec<u8>, bool)>`.

saschagrunert:
Thank you for digging into this, mostly LGTM but the CI does not seem to be happy right now.

haircommander:
huzzah! @saschagrunert @rphillips PTAL

conmon-rs/server/src/attach.rs
```diff
@@ -192,74 +203,96 @@ impl Attach {
...
+                            Err(e) => bail!("unable to write packet {}/{}: {:#}", idx, len, e),
+                        }
                     }
-                    Err(ref e) if e.kind() == ErrorKind::WouldBlock => continue,
-                    Err(ref e) if e.kind() == ErrorKind::BrokenPipe => break,
-                    Err(e) => bail!("unable to write packet {}/{}: {:#}", idx, len, e),
+                }
+                _ = token.cancelled() => {
+                    debug!("Exiting because token cancelled");
+                    match write_half.write(Self::DONE_PACKET).await {
```
rphillips, line 288: should this be a CANCELLED_PACKET ?

haircommander: I don't think we need to make the distinction. either way, the client will disconnect with no error (unless the exit code was non-zero, which is communicated with a different mechanism

conmon-rs/server/src/child_reaper.rs
```diff
@@ -109,6 +111,7 @@ impl ChildReaper {
                     str::from_utf8(&output.stderr).context("convert stderr to utf8")?,
                 )?;
             }
+            token.cancel();
```
rphillips, line 114: this token.cancel() can be skipped due to the previous line. Do you think it should be fixed?

conmon-rs/server/src/streams.rs
```diff
@@ -149,6 +173,8 @@ mod tests {
 
         assert_eq!(msg_string(msg)?, expected);
 
+        token.cancel();
```
rphillips, line 176: hmm. this one can too...

conmon-rs/server/src/child_reaper.rs
haircommander: this is before we've registered the child, so there isn't any entity that will call the cancel. I can add a comment

rphillips:
Nice job! I'll defer to @saschagrunert so he has a chance to look through.

conmon-rs/server/src/streams.rs
haircommander: same as above, the unit test has no entity paying attention to the command, so nothing else would call cancel()

conmon-rs/server/src/attach.rs
```diff
@@ -192,74 +203,96 @@ impl Attach {
...
+        tx: Sender<Vec<u8>>,
+        token: CancellationToken,
+    ) -> Result<()> {
         loop {
             let mut buf = vec![0; Self::PACKET_BUF_SIZE];
-            match read_half.read(&mut buf).await {
-                Ok(n) if n > 0 => {
-                    if let Some(first_zero_idx) = buf.iter().position(|&x| x == 0) {
-                        buf.resize(first_zero_idx, 0);
+            select! {
```
cgwalters, line 213: Beware of `select!` https://blog.yoshuawuyts.com/futures-concurrency-3/

I think in this case it may be OK because we'll never alternate between the two cases - if cancellation token fires, we're done.

But, I started adding "safety" comments to all my uses of `select!` after reading that blog.

conmon-rs/server/src/attach.rs
cgwalters: Oh heh you tagged me earlier and I missed it.

I think this case is safe, but I'd still add a comment with a rationale for why.

saschagrunert (approved):
/lgtm

conmon-rs/server/src/rpc.rs
```diff
@@ -123,13 +124,18 @@ impl conmon::Server for Server {
...
 
-                let grandchild_pid = capnp_err!(match child_reaper
+                let (grandchild_pid, token) = capnp_err!(match child_reaper
                     .create_child(runtime, args, &mut container_io, &pidfile)
                     .await
                 {
                     Err(e) => {
                         // Attach the stderr output to the error message
-                        let (_, stderr, _) = container_io.read_all_with_timeout(None).await;
+                        // TODO FIXME
```
saschagrunert, line 133: What need to be fixed here?

haircommander: ah I had to figure out if creating a new CancellationToken was problematic here. I think it's correct though.

saschagrunert: Closing the loop https://github.com/containers/conmon-rs/pull/699

PR #694, opened 2022-09-08
Close attach readers

### LINKED ISSUE #534: Attach leaks a goroutine
### What happened?

We print the active goroutines after each integration test. It looks like that the `util.CopyDetachable` still leaves a running goroutine open:

```
2022-07-20T14:16:05.7893221Z 2 @ 0x44c076 0x45d112 0x4b882e 0x4b92e5 0x654416 0x6dc739 0x47dfc1
2022-07-20T14:16:05.7893648Z #	0x4b882d	io.(*pipe).read+0x14d										/opt/hostedtoolcache/go/1.18.4/x64/src/io/pipe.go:57
2022-07-20T14:16:05.7894140Z #	0x4b92e4	io.(*PipeReader).Read+0x64									/opt/hostedtoolcache/go/1.18.4/x64/src/io/pipe.go:136
2022-07-20T14:16:05.7895004Z #	0x654415	github.com/containers/common/pkg/util.CopyDetachable+0xb5					/home/runner/go/pkg/mod/github.com/containers/common@v0.48.1-0.20220705175712-dd1c331887b9/pkg/util/copy.go:16
2022-07-20T14:16:05.7895944Z #	0x6dc738	github.com/containers/conmon-rs/pkg/client.(*ConmonClient).setupStdioChannels.func2+0xb8	/home/runner/work/conmon-rs/conmon-rs/pkg/client/attach.go:201
```

https://github.com/containers/conmon-rs/blob/13d4828464b72d0c734643c5ce405dbb60fbb6fe/pkg/client/attach.go#L198-L204

We should find a way to cleanup the goroutine on server shutdown as well.

### What did you expect to happen?

No running goroutines for attach after the integration tests.

### How can we reproduce it (as minimally and precisely as possible)?

Running the integration tests reveals the issue by printing the goroutines at the end.

### Anything else we need to know?

No

### conmon-rs version

<details>

```console
$ conmonrs --version
version: 0.1.0-dev
tag: none
commit: 68252d0373b8e262bdf7ff8780fb2b5ab6f66c29
build: 2022-07-20 14:13:13 +00:00
rustc 1.62.1 (e092d0b6b 2022-07-16)
```

</details>

### OS version

Not relevant

### Additional environment details (AWS, VirtualBox, physical, etc.)

No

saschagrunert:
#### What type of PR is this?

/kind bug

We now close outstanding `Read()` operations on `CopyDetachable` for each attach session. The conmon client collects all attach sessions and also closes them on `Shutdown`.

Fixes #534 
None

```release-note
Cleanup attach sessions to not leak any goroutine.
```

### DIFF
--- a/pkg/client/attach.go
+++ b/pkg/client/attach.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/common/pkg/resize"
 	"github.com/containers/common/pkg/util"
 	"github.com/containers/conmon-rs/internal/proto"
+	"github.com/google/uuid"
 )
 
 const (
@@ -37,8 +38,8 @@ type AttachStreams struct {
 
 // In defines an input stream.
 type In struct {
-	// Wraps an io.Reader
-	io.Reader
+	// Wraps an io.ReadCloser
+	io.ReadCloser
 }
 
 // Out defines an output stream.
@@ -170,14 +171,16 @@ func (c *ConmonClient) attach(ctx context.Context, cfg *AttachConfig) (err error
 		return nil
 	}
 
-	receiveStdoutError, stdinDone := c.setupStdioChannels(cfg, conn)
+	id := uuid.NewString()
+
+	receiveStdoutError, stdinDone := c.setupStdioChannels(cfg, conn, id)
 	if cfg.PostAttachFunc != nil {
 		if err := cfg.PostAttachFunc(); err != nil {
 			return fmt.Errorf("run post attach func: %w", err)
 		}
 	}
 
-	if err := c.readStdio(cfg, conn, receiveStdoutError, stdinDone); err != nil {
+	if err := c.readStdio(cfg, conn, id, receiveStdoutError, stdinDone); err != nil {
 		return fmt.Errorf("read stdio: %w", err)
 	}
 
@@ -185,7 +188,7 @@ func (c *ConmonClient) attach(ctx context.Context, cfg *AttachConfig) (err error
 }
 
 func (c *ConmonClient) setupStdioChannels(
-	cfg *AttachConfig, conn *net.UnixConn,
+	cfg *AttachConfig, conn *net.UnixConn, id string,
 ) (receiveStdoutError, stdinDone chan error) {
 	receiveStdoutError = make(chan error)
 	go func() {
@@ -193,13 +196,18 @@ func (c *ConmonClient) setupStdioChannels(
 	}()
 
 	stdinDone = make(chan error)
-	go func() {
-		var err error
-		if cfg.Streams.Stdin != nil {
-			_, err = util.CopyDetachable(conn, cfg.Streams.Stdin, cfg.DetachKeys)
-		}
-		stdinDone <- err
-	}()
+	stdin := cfg.Streams.Stdin
+
+	if stdin != nil {
+		c.attachReaders.Store(id, stdin)
+
+		go func() {
+			_, err := util.CopyDetachable(conn, stdin, cfg.DetachKeys)
+			if !errors.Is(err, io.ErrClosedPipe) {
+				stdinDone <- err
+			}
+		}()
+	}
 
 	return receiveStdoutError, stdinDone
 }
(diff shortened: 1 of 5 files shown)


pkg/client/attach.go
```diff
@@ -313,6 +343,9 @@ func (c *ConmonClient) readStdio(
 
 	case err = <-stdinDone:
 		c.logger.WithError(err).Trace("Received possible error on input channel")
+
+		c.tryCloseAttachReaderForID(id)
```
haircommander, line 347: hm we send on stdinDone in the CopyDetatchable goroutine. do we need to save the reader if we can close when CopyDetatchable exits?

pkg/client/attach.go
```diff
@@ -170,35 +171,42 @@ func (c *ConmonClient) attach(ctx context.Context, cfg *AttachConfig) (err error
...
 
 	stdinDone = make(chan error)
+	stdin := cfg.Streams.Stdin
+
 	go func() {
-		var err error
-		if cfg.Streams.Stdin != nil {
-			_, err = util.CopyDetachable(conn, cfg.Streams.Stdin, cfg.DetachKeys)
+		if stdin != nil {
+			c.attachReaders.Store(id, stdin)
```
rphillips, line 203: nit... We should reverse the logic and return if stdin == nil, and remove the nesting.

pkg/client/attach.go
```diff
@@ -294,13 +302,35 @@ func (c *ConmonClient) handlePacket(cfg *AttachConfig, buf []byte, nr int) (cont
...
 }
 
+func (c *ConmonClient) tryCloseAttachReaderForID(id string) {
+	c.logger.Tracef("Closing attach reader for ID: %s", id)
+	if val, ok := c.attachReaders.Load(id); ok {
+		c.closeAttachReader(val)
+	}
+}
+
+func (c *ConmonClient) closeAttachReader(val any) {
```
rphillips, line 312: It seems like we can remove any and use a CloseReader, right?

pkg/client/attach.go
saschagrunert: Hm I think we do not close at all in this case, will remove the code path.

pkg/client/attach.go
saschagrunert: Done, changed the logic to not even start the goroutine if `stdin` is `nil`.

pkg/client/attach.go
saschagrunert: The intention was to catch the interface conversion as well because the sync map does not support generics yet.

pkg/client/attach.go
```diff
@@ -170,36 +171,43 @@ func (c *ConmonClient) attach(ctx context.Context, cfg *AttachConfig) (err error
...
+
+	if stdin != nil {
+		c.attachReaders.Store(id, stdin)
+
+		go func() {
+			_, err := util.CopyDetachable(conn, stdin, cfg.DetachKeys)
+			if !errors.Is(err, io.ErrClosedPipe) {
+				stdinDone <- err
+			}
+		}()
```
haircommander, line 209: sorry, my earlier nit was more saying: why don't we call stdin.Close() here, and avoid the `attachReaders` structure all together?

saschagrunert: Because we block in CopyDetachable:

```
1 @ 0x43d936 0x44d7bc 0x497791 0x497f45 0x597c27 0x60b137 0x46ea01
#       0x497790        io.(*pipe).read+0xb0                                                                            /nix/store/mf51hzhs63rjzrp8qqzqpj35j140a9bb-go-1.19/share/go/src/io/pipe.go:57
#       0x497f44        io.(*PipeReader).Read+0x24                                                                      /nix/store/mf51hzhs63rjzrp8qqzqpj35j140a9bb-go-1.19/share/go/src/io/pipe.go:136
#       0x597c26        github.com/containers/common/pkg/util.CopyDetachable+0xa6                                       /home/sascha/go/pkg/mod/github.com/containers/common@v0.49.1/pkg/util/copy.go:16
#       0x60b136        github.com/containers/conmon-rs/pkg/client.(*ConmonClient).setupStdioChannels.func2+0x56        /home/sascha/git/conmon-rs/pkg/client/attach.go:205
```

PR #702, opened 2022-09-12
Fix critest "runtime should support attach" test

saschagrunert:
#### What type of PR is this?

/kind bug

The conformance CRI test does not stop the running container after checking its output. It expects that the connection gets dropped when there is no more input to read. We now adjust the current attach behavior to stop in that case, which hopefully does not break the Podman use case.

See: https://github.com/kubernetes-sigs/cri-tools/blob/909f652/pkg/validate/streaming.go#L266-L305

We also strip down the size of the data packets to return the right amount of received bytes in the client.

A debug message has been changed which correctly logs the amount of written packets.

None
None

```release-note
Fixed critest "runtime should support attach" test.
```

### DIFF
--- a/conmon-rs/server/src/attach.rs
+++ b/conmon-rs/server/src/attach.rs
@@ -103,7 +103,7 @@ impl Attach {
     const PACKET_BUF_SIZE: usize = 8192;
 
     /// The packet indicating that we're done writing.
-    const DONE_PACKET: &'static [u8; Self::PACKET_BUF_SIZE] = &[0; Self::PACKET_BUF_SIZE];
+    const DONE_PACKET: &'static [u8; 1] = &[0];
 
     /// Create a new attach instance.
     fn create<T>(
@@ -242,7 +242,11 @@ impl Attach {
                                 e.raw_os_error().context("get OS error")?
                             ),
                         },
-                        _ => {}
+                        _ => {
+                            debug!("Stopping read loop because there is nothing more to read");
+                            token.cancel();
+                            return Ok(());
+                        }
                     }
                 }
                 _ = token.cancelled() => {
@@ -279,20 +283,19 @@ impl Attach {
                                 Pipe::StdErr => 3,
                             };
                             y.insert(0, p);
-                            y.resize(Self::PACKET_BUF_SIZE, 0);
                             y
                         })
                         .collect::<Vec<_>>();
 
-                    let len = packets.len() - 1;
+                    let len = packets.len();
                     for (idx, packet) in packets.iter().enumerate() {
                         match write_half.write(packet).await {
                             Ok(_) => {
-                                debug!("Wrote {} packet {}/{} to client", pipe, idx, len)
+                                debug!("Wrote {} packet {}/{} to client", pipe, idx + 1, len)
                             }
                             Err(ref e) if e.kind() == ErrorKind::WouldBlock => continue,
                             Err(ref e) if e.kind() == ErrorKind::BrokenPipe => break,
-                            Err(e) => bail!("unable to write packet {}/{}: {:#}", idx, len, e),
+                            Err(e) => bail!("unable to write packet {}/{}: {:#}", idx + 1, len, e),
                         }
                     }
                 }


conmon-rs/server/src/attach.rs
```diff
@@ -237,7 +237,11 @@ impl Attach {
                                 e.raw_os_error().context("get OS error")?
                             ),
                         },
-                        _ => {}
+                        _ => {
+                            debug!("Stopping read loop because there is nothing more to read");
```
haircommander, line 241: I am not sure about this. I think we should double check with a manual check with crictl to make sure it behaves as expected

saschagrunert: Hm, I tested it with CRI-O and it seems to work with crictl. Can we verify if the change also works with Podman somehow?

rphillips: looks like we are missing a match statement when `n==0`

haircommander:
I gave this a test, and the behavior is a bit off:

when running with conmon-rs and cri-o, I do
```
 $ pod_id=$(sudo crictl runp test/testdata/sandbox_config.json )
 $ sudo crictl create $pod_id test/testdata/container_sleep.json test/testdata/sandbox_config.json 
```
in one terminal, then
```
 $ sudo crictl attach $(output from above)
```
in another. Then in the first terminal
```
 $ sudo crictl start $(output from above)
```

This works to attach and we relay the information, but when calling `exit` on the session, it hangs and the terminal doesn't yield to the parent terminal instance. I find this to not be expected, is it supposed to be?

saschagrunert:
> This works to attach and we relay the information, but when calling `exit` on the session, it hangs and the terminal doesn't yield to the parent terminal instance. I find this to not be expected, is it supposed to be?

This also does not work with conmon for me. I'm not sure if this is a bug since I'm not sure if anyone ever described how attach should behave. Do you know if we have any reference to that other than the tests?

haircommander:
> 

hm I am not sure: I think it's working for me. When I call exit within the attached shell with conmon, it exits and the crictl attach call terminates

saschagrunert:
> > 
> 
> hm I am not sure: I think it's working for me. When I call exit within the attached shell with conmon, it exits and the crictl attach call terminates

Yes, you're right. I had to rebase on top of the latest `main` changes to make it work again. Now it runs on my system with the test mentioned in https://github.com/containers/conmon-rs/pull/702#issuecomment-1243880391

May I ask you to verify that too?

conmon-rs/server/src/attach.rs
saschagrunert: Yep, that's what this path is for :+1:

haircommander:
hm it still isn't working for me. the terminal in which I attach still hangs after calling `exit`

saschagrunert:
Hm, I tested it with the follow `Vagrantfile`:

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrant box for testing
Vagrant.configure("2") do |config|
  config.vm.box = "fedora/36-cloud-base"
  memory = 8192
  cpus = 4

  config.vm.provider :virtualbox do |v|
    v.memory = memory
    v.cpus = cpus
  end

  config.vm.provider :libvirt do |v|
    v.memory = memory
    v.cpus = cpus
  end

  config.vm.provision "install-dependencies", type: "shell", run: "once", privileged: false do |sh|
    sh.inline = <<~SHELL
      set -euxo pipefail

      sudo dnf install -y capnproto container-selinux gcc git iptables

      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
      . "$HOME/.cargo/env"

      curl https://storage.googleapis.com/k8s-artifacts-cri-tools/release/v1.25.0/critest-v1.25.0-linux-amd64.tar.gz | sudo tar xfz - -C /usr/bin
      curl https://raw.githubusercontent.com/cri-o/cri-o/v1.25.0/scripts/get | sudo bash -s -- -t v1.25.0

      cat <<EOT | sudo tee /etc/crio/crio.conf.d/10-crun.conf
[crio.runtime]
default_runtime = "runc"

[crio.runtime.runtimes.runc]
runtime_type = "pod"
monitor_path = "/vagrant/target/debug/conmonrs"
EOT

      sudo systemctl enable --now crio.service

      pushd /vagrant
      cargo build
      popd

      git clone --depth=1 https://github.com/cri-o/cri-o
      pushd cri-o

      cat <<EOT | tee ctr.patch
diff --git a/test/testdata/container_sleep.json b/test/testdata/container_sleep.json
index e1bd9868a..c7c85deb9 100644
--- a/test/testdata/container_sleep.json
+++ b/test/testdata/container_sleep.json
@@ -6,10 +6,7 @@
 		"image": "quay.io/crio/redis:alpine"
 	},
 	"command": [
-		  "/bin/sleep", "6000"
-	],
-	"args": [
-                "6000"
+		  "/bin/sh"
 	],
 	"working_dir": "/",
 	"envs": [
@@ -22,9 +19,9 @@
 		"pod": "podsandbox"
 	},
 	"log_path": "",
-	"stdin": false,
+	"stdin": true,
 	"stdin_once": false,
-	"tty": false,
+	"tty": true,
 	"linux": {
 		"security_context": {
 			"namespace_options": {
EOT

      git apply ctr.patch

      sudo crictl pull quay.io/crio/redis:alpine
      pod_id=$(sudo crictl runp test/testdata/sandbox_config.json)
      sudo crictl create $pod_id test/testdata/container_sleep.json test/testdata/sandbox_config.json >> /tmp/ctr
    SHELL
  end
end
```

And it seems to work:

First terminal:
```console
> vagrant ssh
Last login: Wed Sep 14 12:23:40 2022 from 192.168.121.1
[vagrant@fedora ~]$ sudo crictl attach -it $(cat /tmp/ctr)
/ # exit
[vagrant@fedora ~]$
```

Second terminal:
```console
> vagrant ssh
[vagrant@fedora ~]$ sudo crictl start $(cat /tmp/ctr)
a413fd949113d383cfd46950ad69aabc541b13120cb0a47fb443234efecddbad
[vagrant@fedora ~]$
```

@haircommander do you see any significant difference compared to your test environment?

haircommander:
huh it works now :magic_wand: 
/lgtm

PR #703, opened 2022-09-12
Add critest run to CI

saschagrunert:
#### What type of PR is this?

/kind ci

We now run a critest suite with a pinned CRI-O release for each PR.

None

Needs https://github.com/containers/conmon-rs/pull/702 to not timeout during the attach test.

```release-note
None
```

### DIFF
--- a/.github/setup
+++ b/.github/setup
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+set -euox pipefail
+
+main() {
+    set -x
+    prepare_system
+
+    install_critest
+    install_crio
+}
+
+curl_retry() {
+    curl -sSfL --retry 5 --retry-delay 3 "$@"
+}
+
+prepare_system() {
+    sudo systemctl stop docker
+    sudo ufw disable
+
+    # enable necessary kernel modules
+    sudo ip6tables --list >/dev/null
+
+    # enable necessary sysctls
+    sudo sysctl -w net.ipv4.conf.all.route_localnet=1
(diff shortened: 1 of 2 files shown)


.github/install-cri-o
```diff
@@ -0,0 +1,71 @@
...
+    if ! grep -q containers /etc/subgid; then
+        echo "containers:100000:65536" | sudo tee -a /etc/subgid
+    fi
+
+    printf "RateLimitInterval=0\nRateLimitBurst=0\n" | sudo tee /etc/systemd/journald.conf
+    sudo systemctl restart systemd-journald
+}
+
+install_crio() {
+    TAG=v1.25.0
```
rphillips, line 43: Could we fetch the latest tag?

conmon-rs/server/src/attach.rs
```diff
@@ -237,7 +237,11 @@ impl Attach {
                                 e.raw_os_error().context("get OS error")?
                             ),
                         },
-                        _ => {}
+                        _ => {
```
rphillips, line 240: We are missing the n == 0 case. Being explicit with that case should be able to replace the '_' case.

rphillips: Ah, I see your note on the other PR

.github/install-cri-o
haircommander: even better, we could just use `main` here

saschagrunert: Alright, changing it to `main`.

PR #704, opened 2022-09-12
Reduce memory usage for attach

saschagrunert:
#### What type of PR is this?

/kind cleanup

It's unrealistic that we have to buffer 1000 packets, especially since they should be consumed by in a fast manner. We now still keep a safe zone of 4 packets in the channel, but this already reduces the memory by multiple hundred kilobytes.

We also optimize the release build by stripping of the binary as well as removing the debug symbols per default.

None
None

```release-note
None
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,9 @@ members = [
 ]
 
 [profile.release]
+debug = 2
 lto = true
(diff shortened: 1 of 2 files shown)


haircommander:
nice!
/lgtm

Issue #720, opened 2022-09-16
Kubernetes e2e test `Kubectl Port forwarding With a server listening on localhost that expects a client request should support a client that connects, sends DATA, and disconnects` fails


saschagrunert:
### What happened?

The test fails in https://github.com/cri-o/cri-o/pull/6220 together with other cases, but I assume that the root cause is the same.

Interestingly, on my local machine the cancellation of the execsync token seems to kill the main container PID.

### What did you expect to happen?

That the test succeeds.

#### Running the test
```
> k8s-test-run --ginkgo.focus "Kubectl Port forwarding With a server listening on localhost that expects a client request should support a client that connects, sends DATA, and disconnects"
…
[BeforeEach] [sig-cli] Kubectl Port forwarding
  set up framework | framework.go:158
STEP: Creating a kubernetes client 09/16/22 11:08:56.88
Sep 16 11:08:56.880: INFO: >>> kubeConfig: /var/run/kubernetes/admin.kubeconfig
STEP: Building a namespace api object, basename port-forwarding 09/16/22 11:08:56.881
STEP: Waiting for a default service account to be provisioned in namespace 09/16/22 11:08:56.885
STEP: Waiting for kube-root-ca.crt to be provisioned in namespace 09/16/22 11:08:56.887
[It] should support a client that connects, sends DATA, and disconnects
  test/e2e/kubectl/portforward.go:481
STEP: Creating the target pod 09/16/22 11:08:56.888
Sep 16 11:08:56.892: INFO: Waiting up to 5m0s for pod "pfpod" in namespace "port-forwarding-8233" to be "running and ready"
Sep 16 11:08:56.893: INFO: Pod "pfpod": Phase="Pending", Reason="", readiness=false. Elapsed: 1.353931ms
Sep 16 11:08:56.893: INFO: The phase of Pod pfpod is Pending, waiting for it to be Running (with Ready = true)
Sep 16 11:08:58.896: INFO: Pod "pfpod": Phase="Running", Reason="", readiness=false. Elapsed: 2.003699716s
Sep 16 11:08:58.896: INFO: The phase of Pod pfpod is Running (Ready = false)
Sep 16 11:09:00.896: INFO: Pod "pfpod": Phase="Running", Reason="", readiness=false. Elapsed: 4.003966771s
Sep 16 11:09:00.896: INFO: The phase of Pod pfpod is Running (Ready = false)
[...]

PR #734, opened 2022-09-20
Send only streams data if channel is not closed

saschagrunert:
#### What type of PR is this?

/kind bug

The stream data is only read on `read_all_with_timeout`, which is called via exec sync or on on container creation error. This means we can end-up having a closed sender, but we should not fail in this case because it will lead into unexpected container termination.

Fixes https://github.com/containers/conmon-rs/issues/720

None

```release-note
Fixed bug where a closed stream sender causes container termination.
```

### DIFF
--- a/conmon-rs/server/src/container_io.rs
+++ b/conmon-rs/server/src/container_io.rs
@@ -267,9 +267,11 @@ impl ContainerIO {
                                 .await
                                 .context("write to attach endpoints")?;
 
-                            message_tx
-                                .send(Message::Data(data.into(), pipe))
-                                .context("send data message")?;
+                            if !message_tx.is_closed() {
+                                message_tx
+                                    .send(Message::Data(data.into(), pipe))
+                                    .context("send data message")?;
+                            }
                         }
                         Err(e) => match Errno::from_i32(e.raw_os_error().context("get OS error")?) {
                             Errno::EIO => {


saschagrunert:
An alternative would be to read after successful container creation:

```dif
diff --git a/conmon-rs/server/src/container_io.rs b/conmon-rs/server/src/container_io.rs
index b1ebc52..c1e4a61 100644
--- a/conmon-rs/server/src/container_io.rs
+++ b/conmon-rs/server/src/container_io.rs
@@ -267,11 +267,9 @@ impl ContainerIO {
                                 .await
                                 .context("write to attach endpoints")?;
 
-                            if !message_tx.is_closed() {
-                                message_tx
-                                    .send(Message::Data(data.into(), pipe))
-                                    .context("send data message")?;
-                            }
+                            message_tx
+                                .send(Message::Data(data.into(), pipe))
+                                .context("send data message")?;
                         }
                         Err(e) => match Errno::from_i32(e.raw_os_error().context("get OS error")?) {
                             Errno::EIO => {
index daea94e..5462ced 100644
@@ -14,7 +14,7 @@ use std::{
     process, str,
     time::Duration,
 };
-use tokio::time::Instant;
+use tokio::{task, time::Instant};
 use tracing::{debug, debug_span, error, Instrument};
 use uuid::Uuid;
 
@@ -150,18 +150,28 @@ impl conmon::Server for Server {
 
                 // register grandchild with server
                 let io = SharedContainerIO::new(container_io);
+                let io_clone = io.clone();
                 let child = Child::new(
                     id,
                     grandchild_pid,
                     exit_paths,
                     oom_exit_paths,
                     None,
-                    io,
+                    io_clone,
                     cleanup_cmd,
                     token,
                 );
                 capnp_err!(child_reaper.watch_grandchild(child))?;
 
+                task::spawn(
+                    async move {
+                        if let Err(e) = io.read_all_with_timeout(None).await {
+                            error!("IO read failure: {:#}", e);
+                        }
+                    }
+                    .instrument(debug_span!("read_all_with_timeout")),
+                );
+
                 results
                     .get()
                     .init_response()

```

saschagrunert:
So the e2e tests are now looking way better with this change: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/6220/pull-ci-cri-o-cri-o-main-ci-e2e-conmonrs/1572188830188965888

The 2 failing tests are also failing on `main` right now.

haircommander:
/lgtm

good find!

Issue #740, opened 2022-09-21
Allow CRI-O logging to stdout


saschagrunert:
We right now use the default logger in CRI-O, which logs to systemd. It would be good to have at least a runtime monitor configuration to also log to stdout. This allows us to collect conmon and CRI-O logs in parallel.

Q: Do we wanna make stdout logging the default? Which benefits do we gain when logging to systemd?

haircommander:
can we stay attached to stdout if we've daemonized? conmon-rs is a child of systemd, not cri-o, and we may lose the logs (we definitely would if cri-o restarts, but may unconditionally)

saschagrunert:
Hm, setting the log driver to stdout works when CRI-O runs in foreground in my terminal, then both logs appear side by side. A CRI-O restart would probably loose that connection, but we could establish an IO reconnect on restore (or similar).  🤔

saschagrunert:
I think we can use multiple logging drivers in parallel and add a file-based logger later on. This can be used by CRI-O to stream the logs.

haircommander:
what do you think about enabling verbose logs in the journal? I am open to a file based log driver, but it feels more idiomatic (and closer to conmon's behavior) to have it logged to journal

saschagrunert:
> what do you think about enabling verbose logs in the journal? I am open to a file based log driver, but it feels more idiomatic (and closer to conmon's behavior) to have it logged to journal

I'm wondering how this should work, for example `sudo journalctl _COMM=conmonrs -o verbose` outputs the following format:

```
Tue 2022-09-20 10:53:47.707490 CEST [s=3ec34f7a8bd34c4e97b1f8702442e55e;i=123a55b;b=8f6c219633ce46e3a1c4dbad7043bc47;m=1aeb85159;t=5e917f765a89a;x=73e177f314cf05b0]
    _UID=0
    _AUDIT_LOGINUID=1000
    _AUDIT_SESSION=1
    _SELINUX_CONTEXT=kernel
    _BOOT_ID=8f6c219633ce46e3a1c4dbad7043bc47
    _MACHINE_ID=4ef6243ad0d84d968b0a6e3060f537ba
    _HOSTNAME=nixos
    PRIORITY=6
    _GID=0
    _CAP_EFFECTIVE=1ffffffffff
    _SYSTEMD_CGROUP=/user.slice/user-1000.slice/session-1.scope
    _SYSTEMD_SESSION=1
    _SYSTEMD_OWNER_UID=1000
[...]

PR #770, opened 2022-10-04
Allow multiple log drivers to be specified

saschagrunert:
#### What type of PR is this?

/kind api-change

This extends the API to allow multiple log drivers in parallel. We can later on support other drivers like files to log into multiple destinations.

Part of #740 
None

```release-note
Changed `--log-driver` to `--log-drivers` to support multiple server logging drivers simultaneously.
```


conmon-rs/server/src/config.rs
```diff
@@ -48,17 +48,18 @@ pub struct Config {
...
+        long("log-drivers"),
         short('d'),
         possible_values(LogDriver::iter().map(|x| x.into()).collect::<Vec<&str>>()),
-        value_name("DRIVER")
+        value_name("DRIVERS")
     )]
-    /// The logging driver used by the conmon server.
-    log_driver: LogDriver,
+    /// The logging drivers used by the server. Can be specified multiple times.
+    log_drivers: Vec<LogDriver>,
```
haircommander, line 62: unfortunately, I think we need to keep both. We can drop `log_driver` when conmon-rs moves to 1.0

Issue #796, opened 2022-10-12
Propagate Open Telemetry context from CRI-O to conmon-rs


saschagrunert:
We can use the Open Telemetry context propagation to link spans between CRI-O and conmon-rs. This would mean that we have to use the global Text Map Propagator on the client side as well as on the conmon-rs server side to inject/extract the spans.

Since capnproto does neither natively support serializing hash maps nor carrying headers, we have to change each RPC to additionally contain the metadata:

```
struct VersionRequest {
    metadata @0 :Data;
    verbose @1 :Bool;
}
```

I assume that we introduce a new metadata type like:

```rust
struct MetadataMap<'a>(HashMap<&'a str, &'a str>);

impl<'a> Extractor for MetadataMap<'a> {
    fn get(&self, key: &str) -> Option<&str> {
        self.0.get(key).map(|x| x.clone())
    }

    /// Collect all the keys from the MetadataMap.
    fn keys(&self) -> Vec<&str> {
        self.0.keys().map(|k| k.clone()).collect::<Vec<_>>()
    }
}
```

And then (for each RPC) restore the spans:

```rust
let metadata = pry!(req.get_metadata());
let hashmap: HashMap<&str, &str> = pry_err!(serde_json::from_slice(metadata));
let metadata_map = MetadataMap(hashmap);
let parent_cx = global::get_text_map_propagator(|prop| prop.extract(&metadata_map));
self.tracer().start_with_context("version", &parent_cx);
```

Examples:
- https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/grpc/src/server.rs
- https://github.com/open-telemetry/opentelemetry-go/blob/main/example/passthrough/handler/handler.go

PR #799, opened 2022-10-12
Propagate existing telemetry spans to server

saschagrunert:
#### What type of PR is this?

/kind feature

Existing telemetry spans will be now propagated to conmon-rs.

Fixes https://github.com/containers/conmon-rs/issues/796
Example
screenshot

```release-note
Propagate existing Open Telemetry spans to the server, for example when being used via CRI-O.
```


conmon-rs/common/proto/conmon.capnp
```diff
@@ -4,7 +4,8 @@ interface Conmon {
     ###############################################
     # Version
     struct VersionRequest {
-        verbose @0 :Bool;
+        metadata @0 :Data;
```
haircommander, line 7: when adding new fields, @ number should be higher so we have backwards compatibility. the order of the fields don't matter, so this could be
```
metadata @1 :Data;
verbose @0 :Bool;
```

saschagrunert: Makes sense, I'll change it that way.

saschagrunert:
@haircommander @rphillips PTAL, this is now ready for review.

pkg/client/client.go
```diff
@@ -456,6 +460,14 @@ func (c *ConmonClient) Version(
 			return fmt.Errorf("create request: %w", err)
 		}
 
+		metadata, err := c.metadataBytes(ctx)
```
haircommander, line 463: should we only do these if we've previously EnabledTracing? We could save it in the client

conmon-rs/common/proto/conmon.capnp
```diff
@@ -4,10 +4,12 @@ interface Conmon {
     ###############################################
     # Version
     struct VersionRequest {
+        metadata @1 :Data;
```
rphillips, line 7: nit... I think we should preserve the ordering within the file on all the added metadata, so it's easier to add new fields later.

pkg/client/client.go
saschagrunert: Yes, making it a no-op when tracing is disabled.

conmon-rs/common/proto/conmon.capnp
saschagrunert: Fixed as suggested

saschagrunert:
@haircommander @rphillips PTAL again

rphillips:
lgtm 

@haircommander ?

PR #804, opened 2022-10-14
Shutdown tracing after runtime

saschagrunert:
#### What type of PR is this?

/kind bug

This makes the overall exporter race a bit more resilient against race conditions.

None
We still seem to miss the root spans in some cases, but the overall result for a container creation will now look like this, which is promising:

screenshot

```release-note
None
```


saschagrunert:
Ugh, that label add wasn’t intentional. 🫣

haircommander:
it's okay, I meant to lgtm :)

PR #822, opened 2022-10-20
Allow client traces by passing a tracer

saschagrunert:
#### What type of PR is this?

/kind feature

If we now pass down a `trace.Tracer` via the configuration, conmonrs is able to add additional spans based on how the client is being used.

None
None

```release-note
Added `Tracer trace.Tracer` to the server configuration, which allows the client to create additional spans.
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -47,6 +47,7 @@ type ConmonClient struct {
 	logger         *logrus.Logger
 	attachReaders  *sync.Map // K: UUID string, V: *attachReaderValue
 	tracingEnabled bool
+	tracer         trace.Tracer
 }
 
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -236,6 +267,11 @@ func (c *ConmonClient) startServer(config *ConmonServerConfig) error {
 }
 
 func (c *ConmonClient) toArgs(config *ConmonServerConfig) (entrypoint string, args []string, err error) {
+	_, span := c.startSpan(context.TODO(), "toArgs")
```
haircommander, line 270: tbh, I don't see how useful inspecting this span would be. I would vote it should be dropped

saschagrunert: Good, I removed that span.

PR #898, opened 2022-11-09
rust: Port client to use tokio too

cgwalters:
I was looking at this code as an example of Rust + capnp, and was a bit surprised at the usage of the non-tokio networking crate.

In digging some, the server is already using the tokio compat bridges, so let's do the same on the client.

This drops out many duplicative dependencies.


rphillips:
Nice. Thanks!

/lgtm

Issue #871, opened 2022-11-04
Experiment moving pinns functionality into conmonrs


haircommander:
[pinns](https://github.com/cri-o/cri-o/tree/main/pinns) is a utility called by cri-o to create pod-level namespaces. It would be cool if conmon-rs could create those namespaces for the pod. Most notably, if it could manage a `pause` process to hold open the pod's PID namespace, then CRI-O could get rid of the infra container

haircommander:
@utam0k I know you were looking for some work in conmon-rs and this seems like something that could be up your alley. do you have any interest in investigating?

utam0k:
> @utam0k I know you were looking for some work in conmon-rs and this seems like something that could be up your alley. do you have any interest in investigating?

Thanks for your kind mention :pray: I'd love to, but I already have some of my hobby tasks, such as KubeDay Japan and runc's issue... So I feel I can't find the time to do that this year. If this issue doesn't need to achieve this year, I'd love to do it next year!

haircommander:
yeah no rush, we can sync in the future :)

PR #908, opened 2022-11-17
Add support for `pause`

saschagrunert:
#### What type of PR is this?

/kind feature

This experiment adds support for running a global shared pause instance side by side to `conmonrs`. The idea is that this instance keeps the required namespaces open and provides them as part of the container creation response to the consumers.

Refers to https://github.com/containers/conmon-rs/issues/871

None

```release-note
TBD
```

### DIFF
--- a/conmon-rs/common/proto/conmon.capnp
+++ b/conmon-rs/common/proto/conmon.capnp
@@ -120,4 +120,31 @@ interface Conmon {
     }
 
     setWindowSizeContainer @5 (request: SetWindowSizeRequest) -> (response: SetWindowSizeResponse);
+
+    ###############################################
+    # CreateNamespaces
+    struct CreateNamespacesRequest {
+        metadata @0 :Data; # Standard metadata to carry.
+        namespaces @1 :List(Namespace); # The list of namespaces to unshare.
+    }
+
+    enum Namespace {
+        ipc @0; # Unshare the IPC namespace.
+        net @1; # Unshare the network namespace.
+        pid @2; # Unshare the PID namespace.
+        user @3; # Unshare the user namespace.
+        uts @4; # Unshare the UTS namespace.
+    }
+
+    struct CreateNamespacesResponse {
+        namespaces @0 :List(NamespaceResponse);  # The list of created namespaces.
+    }
+
+    # Available namespaces.
+    struct NamespaceResponse {
+        type @0 :Namespace; # The type of the namespace.
+        path @1 :Text; # Path to the directory for the unshared namespaces.
+    }
+
+    createNamespaces @6 (request: CreateNamespacesRequest) -> (response: CreateNamespacesResponse);
 }
--- a/conmon-rs/server/src/config.rs
+++ b/conmon-rs/server/src/config.rs
@@ -1,6 +1,6 @@
 //! Configuration related structures
 use anyhow::{bail, Result};
-use clap::{ArgEnum, Parser};
+use clap::{ArgEnum, Parser, Subcommand};
 use getset::{CopyGetters, Getters, Setters};
 use serde::{Deserialize, Serialize};
 use std::{fs, path::PathBuf};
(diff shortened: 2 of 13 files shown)


saschagrunert:
@haircommander @rphillips do you think this would work? The main idea is that conmonrs calls it self via `conmonrs -p` to have something similar like pinns. Now we also forkt to have a new PIDNS at hand. At some later implementation step we can then inherit the namespaces by the pause PID…

conmon-rs/common/proto/conmon.capnp
```diff
@@ -120,4 +121,16 @@ interface Conmon {
...
     setWindowSizeContainer @5 (request: SetWindowSizeRequest) -> (response: SetWindowSizeResponse);
+
+    ###############################################
+    # CreateNamespaces
+    struct CreateNamespacesRequest {
+        metadata @0 :Data;
+    }
+
+    struct CreateNamespacesResponse {
+        path @0 :Text;
```
saschagrunert, line 132: Is this enough or should we abstract even more? Right now we can expect mounted namespaces following the naming convention under this path. We could also have dedicated types mapped around it.

conmon-rs/server/src/config.rs
```diff
@@ -33,6 +33,11 @@ pub struct Config {
     /// Show version information, specify "full" for verbose output.
     version: Option<Verbosity>,
 
+    #[get = "pub"]
+    #[clap(long("pause"), short('p'), value_name("PATH"))]
+    /// Run the pause process instead of the server.
+    pause: Option<PathBuf>,
```
rphillips, line 39: Sorta thinking this functionality should be a subcommand of conmonrs... ie: `conmonrs pause --path=...`. Thoughts?

saschagrunert: Yep, done :+1:

saschagrunert:
Ready for review. PTAL @rphillips @haircommander

conmon-rs/server/src/pause.rs
```diff
@@ -0,0 +1,265 @@
...
+        if pid {
+            flags.insert(CloneFlags::CLONE_NEWPID);
+            namespaces.push(Namespace::Pid);
+        }
+        if net {
+            flags.insert(CloneFlags::CLONE_NEWNET);
+            namespaces.push(Namespace::Net);
+        }
+        if user {
+            // flags.insert(CloneFlags::CLONE_NEWUSER);
```
haircommander, line 169: left over? if not, I think we could use a comment

haircommander:
if conmon-rs exits, are we making sure the pause processes are stopped? I think otherwise they'd leak

conmon-rs/server/src/pause.rs
saschagrunert: Yep, added a comment.

saschagrunert:
> if conmon-rs exits, are we making sure the pause processes are stopped? I think otherwise they'd leak

Yes, there: https://github.com/containers/conmon-rs/pull/908/files#diff-5d1350be7d73e8dfd3f43f70d9d71838500a7c04b2ee5ee19f1c4521f9df7d12R224-R226

conmon-rs/server/src/pause.rs
saschagrunert: We can add support for userns at the next iteration before releasing 0.5.0. Should be straightforward to implement.

saschagrunert:
Gave it a rebase.

rphillips:
This looks ready to me. Nice job... @haircommander ?

haircommander:
well done here!

conmon-rs/server/src/pause.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+/// The global shared multiple pause instance.
+static PAUSE: OnceCell<Pause> = OnceCell::new();
+
+/// The global path for storing bin mounted namespaces.
+const PAUSE_PATH: &str = "/var/run/conmonrs";
+
+/// The file path for storing the pause PID.
+const PAUSE_PID_FILE: &str = ".pause_pid";
```
cgwalters, line 45: Would it make sense to run this as a systemd unit instead?

saschagrunert: Hm, I'm not sure about the resource overhead we add when using systemd here. We can definitely give it a try and see how it behaves.

PR #926, opened 2022-11-28
WIP: Implement ebpf based PID watcher

saschagrunert:
#### What type of PR is this?

/kind feature

An experimental PID watcher using ebpf.

None

None

```release-note
Added ebpf based PID watcher.
```


rphillips:
woah. nice

haircommander:
have you thought about using `aya-rs`? I found the rust integration to be simpler but the documentation was a bit tough to understand

saschagrunert:
> have you thought about using `aya-rs`? I found the rust integration to be simpler but the documentation was a bit tough to understand

Not until now,  but we can elaborate on that.

Makefile
```diff
@@ -19,7 +19,7 @@ release:
 
 .PHONY: release-static
 release-static:
-	RUSTFLAGS="-C target-feature=+crt-static" cargo build --release --target x86_64-unknown-linux-gnu
+	RUSTFLAGS="-C target-feature=+crt-static -L /usr/lib/x86_64-linux-gnu" cargo build -v --release --target x86_64-unknown-linux-gnu --features static
```
saschagrunert, line 22: `-L /usr/lib/x86_64-linux-gnu` is not passed down to the `libbpf-sys` crate, which makes it fail to find libelf as well as libz.

saschagrunert: Running out of ideas, opened https://github.com/libbpf/libbpf-sys/issues/49

mgjm:
@saschagrunert Can you briefly explain to me what the benefits of using ebpf are compared to the existing approach of waitpid and cgroup based oom_watcher?

I know that the current implementation of waitpid does not scale that well (blocking thread per child) and that the native implementation in tokio also does not scale that well (SIGCHLD delivered to **all** waiting childs). But we could implement a single SIGCHLD handler that dispatches the signal to the correct child task.

saschagrunert:
@mgjm this was mostly an experiment to have a more resilient OOM detection. Sometimes we miss OOMs based on the cgroup watching (revealed in our tests). The PID watcher was more an additional nice-to-have here.

Issue #736, opened 2022-09-20
`[sig-cli] Kubectl client Simple pod should support inline execution and attach` fails


saschagrunert:
### Which jobs are failing?

CRI-O conmonrs e2e feature tests.

### Which tests are failing?

Mentioned in the title, ran via: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/6220/pull-ci-cri-o-cri-o-main-ci-e2e-conmonrs/1572188830188965888

### Since when has it been failing?

Since ever, reproducible with 8e4cec9644f742be8334ea8ecfc9f895b1d6691e

### Reason for failure (if possible)

I guess it’s something in the attach stdin code.

saschagrunert:
https://github.com/containers/conmon-rs/pull/746 partly adresses this one.

The `--leave-stdin-open=true` tests are still failing, but those have to be fixed in another PR:

https://github.com/kubernetes/kubernetes/blob/7f129f1c9af62cc3cd4f6b754dacdf5932f39d5c/test/e2e/kubectl/kubectl.go#L614-L648

saschagrunert:
Latest run: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/cri-o_cri-o/6220/pull-ci-cri-o-cri-o-main-ci-e2e-conmonrs/1572942729908850688

PR #991, opened 2023-01-10
Drain stdin on attach

saschagrunert:
#### What type of PR is this?

/kind failing-test

The Kubernetes e2e test flakes because of a token cancellation race between stdin and attach read:

```
[FAIL] [sig-cli] Kubectl client Simple pod [It] should support inline execution and attach
…
  << Timeline

  [FAILED] Expected
      <string>: read:stdin closed

  to contain substring
      <string>: read:value
  In [It] at: test/e2e/kubectl/kubectl.go:764 @ 01/04/23 15:09:48.618
```
https://github.com/kubernetes/kubernetes/blob/5cbd6960c805c22e1c741dde3ce9f2e10f869bce/test/e2e/kubectl/kubectl.go#L757-L766

The issue is that we cancel the token immediately which stops reading from attach. But the container command `echo -n read: && cat && echo 'stdin closed'` within the test provides additional data, which gets no time to be delivered.

We now drain stdin accordingly and wait for all data to be passed down to the receiver.

Refers to https://github.com/containers/conmon-rs/issues/736
None

```release-note
Correctly drain stdin on attach.
```

### DIFF
--- a/conmon-rs/server/src/attach.rs
+++ b/conmon-rs/server/src/attach.rs
@@ -90,13 +90,6 @@ impl SharedContainerAttach {
             .context("receive attach message")
     }
 
-    /// Try to read from all attach endpoints standard input and return the first result.
-    pub fn try_read(&mut self) -> Result<Vec<u8>> {
-        self.read_half_rx
-            .try_recv()
-            .context("try to receive attach message")
-    }
-
     /// Write a buffer to all attach endpoints.
     pub async fn write(&mut self, m: Message) -> Result<()> {
         if self.write_half_tx.receiver_count() > 0 {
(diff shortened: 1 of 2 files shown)


sohankunkerkar (approved):
LGTM

cgwalters:
Doesn't this just make flakes less likely?  I am not familiar with all the details here, just a general suspicion of fixing race conditions with `sleep()`.  Offhand wouldn't we need to read from the fd to end?

saschagrunert:
> Doesn't this just make flakes less likely? 

That's true. One way to fix it would be to increase the complexity by having cancellation tokens for input and output streams. I can follow-up on that, but I'm not sure if the effort is comparable to the test fix.

saschagrunert:
I think we can simply `attach.read()` and wait blocking because we already drop the read half in case of EOF. PTAL again @rphillips @haircommander @cgwalters

rphillips:
lgtm

haircommander:
/lgtm

good idea!

PR #1004, opened 2023-01-12
Revert "Drain stdin on attach"

saschagrunert:
Reverts containers/conmon-rs#991

This will not fix the problem and will make it even worse. We have to somehow split-up the cancellation tokens to wait for stdin and stdout/err separately.


saschagrunert:
PTAL @rphillips @haircommander

PR #1064, opened 2023-01-31
Bind missing namespaces to host

saschagrunert:
#### What type of PR is this?

/kind bug

We now bind all namespaces, independently if unshared or not. If a namespace is not being unshared, then we fallback to the host.

Required for https://github.com/cri-o/cri-o/pull/6590
None

```release-note
Fixed pinning all namespaces to the host when not selected for unsharing in `conmonrs pause`
```


saschagrunert:

PR #1073, opened 2023-02-01
Use zeitgeist for dependency verification

saschagrunert:
#### What type of PR is this?

/kind ci

We now use a dedicated job for verifying our dependencies while pinning them in `dependencies.yaml`.

None

None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ PACKAGE_NAME ?= $(shell cargo metadata --no-deps --format-version 1 | jq -r '.pa
 PREFIX ?= /usr
 CI_TAG ?=
 GOLANGCI_LINT_VERSION := v1.50.1
+ZEITGEIST_VERSION := v0.3.5
+GINKGO_VERSION := v2.8.0
 
 default:
 	cargo build
@@ -35,6 +37,10 @@ lint-go: .install.golangci-lint
 	$(GOTOOLS_BINDIR)/golangci-lint linters
 	GL_DEBUG=gocritic $(GOTOOLS_BINDIR)/golangci-lint run
 
+.PHONY: verify-dependencies
+verify-dependencies: $(GOTOOLS_BINDIR)/zeitgeist
+	$(GOTOOLS_BINDIR)/zeitgeist validate --local-only --base-path . --config dependencies.yaml
+
 unit:
 	cargo test --no-fail-fast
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -54,12 +60,18 @@ integration-static: .install.ginkgo # It needs to be release so we correctly tes
 	sudo -E "$(GOTOOLS_BINDIR)/ginkgo" $(TEST_FLAGS) $(GINKGO_FLAGS)
 
 .install.ginkgo:
-	GOBIN=$(abspath $(GOTOOLS_BINDIR)) go install github.com/onsi/ginkgo/v2/ginkgo@latest
+	GOBIN=$(abspath $(GOTOOLS_BINDIR)) go install github.com/onsi/ginkgo/v2/ginkgo@v$(GINKGO_VERSION)
```
rphillips, line 63: Let's move the 'v' into the GINKGO_VERSION, so there isn't any 'magic'.

Makefile
```diff
@@ -54,12 +60,18 @@ integration-static: .install.ginkgo # It needs to be release so we correctly tes
...
 
 .install.golangci-lint:
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
 		BINDIR=$(abspath $(GOTOOLS_BINDIR)) sh -s $(GOLANGCI_LINT_VERSION)
 
+$(GOTOOLS_BINDIR)/zeitgeist:
+	mkdir -p $(GOTOOLS_BINDIR)
+	curl -sSfL -o $(GOTOOLS_BINDIR)/zeitgeist \
+		https://github.com/kubernetes-sigs/zeitgeist/releases/download/v$(ZEITGEIST_VERSION)/zeitgeist_$(ZEITGEIST_VERSION)_linux_amd64
```
rphillips, line 72: ditto

dependencies.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - name: go
+    version: 1.19
+    refPaths:
+      - path: .github/workflows/ci.yml
+        match: GO_VERSION
+      - path: .github/workflows/release.yml
+        match: GO_VERSION
+
+  - name: rust
+    version: 1.58.1
```
rphillips, line 12: Can we use a newer version of rust?

Makefile
saschagrunert: Fixed :+1:

Makefile
saschagrunert: Done :+1:

dependencies.yaml
saschagrunert: I don't think so. Maybe @lsm5 has more input on when to expect new rust versions on RHEL / RHCOS 8/9.

PR #1082, opened 2023-02-03
Update golangci-lint and config

saschagrunert:
#### What type of PR is this?

/kind ci

Update golangci lint to the latest version and enable linters.

None
None

```release-note
None
```


.golangci.yml
```diff
@@ -91,6 +93,7 @@ linters:
     - whitespace
     - wrapcheck
     # - exhaustruct
+    # - ginkgolinter
```
saschagrunert, line 96: Seems to crash, we can try to enable it again at some later point in time.

Issue #1066, opened 2023-01-31
pause process should only be started if pid namespace is specified


haircommander:
It's possible we do this now, but the pause process should only be created and tracked if the pod has a pod level pid namespace. Otherwise, we're needlessly creating the infra container again (with less overhead, but still).

saschagrunert:
Should we handle that in CRI-O?

saschagrunert:
@haircommander in CRI-O, the check would look like:
```go
sb.NamespaceOptions().GetPid() == types.NamespaceMode_POD
```
Should we pass the `sb.NamespaceOptions().GetPid() ` to the conmon-rs client and no-op if the above condition is met? Or how should that look like interface wise?

haircommander:
conmon-rs already knows what the pid namespace will be because it's created if the user specifies pid in the RPC call. Even if we're bind mounting the pid namespace from the host, there's already a process holding it open (systemd) so we don't need to spawn a pause process. We only need the pause process if the user requested a pod level pid namespace

PR #1085, opened 2023-02-06
Add dedicated error if no PID namespace should be unshared

saschagrunert:
#### What type of PR is this?

/kind cleanup

We do not have to create the pause process on `CreateNamespaces` if no PID namespace should be unshared. In this case we now return a dedicated error and let the users decide what to do with it.

Fixes https://github.com/containers/conmon-rs/issues/1066

None

```release-note
Return `ErrNoPIDNamespaceSpecified` if no PID namespace should be unshared in the `CreateNamespaces` method.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -100,16 +100,16 @@ linters:
     # - wsl
 linters-settings:
   funlen:
-    lines: 155
+    lines: 200
     statements: 50
   varnamelen:
     min-name-length: 1
   cyclop:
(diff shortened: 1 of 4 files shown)


saschagrunert:

haircommander:
I feel unsure about this approach... I still like the idea of having conmon-rs create all the namespaces for a pod eventually and deprecating pinns

saschagrunert:
Yeah I see.

PR #1108, opened 2023-02-14
Enable packit for `main` branch

saschagrunert:
#### What type of PR is this?

/kind ci

Enable packit builds for the main branch.

None
PTAL @lsm5 

```release-note
None
```

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -8,7 +8,8 @@
 specfile_path: conmon-rs.spec
 
 jobs:
-  - job: copr_build
+  - &copr
+    job: copr_build
     trigger: pull_request
     owner: rhcontainerbot
     project: packit-builds
@@ -22,3 +23,7 @@ jobs:
         - "rpkg spec --outdir ./"
       fix-spec-file:
         - "bash .packit.sh"
+
+  - <<: *copr
+    trigger: commit
+    branch: main


saschagrunert:
> @saschagrunert I'm cool with this but do we actually need the `trigger: commit` ? IIUC, it triggers builds after the commit is done, which is already being done by the webhook on `rhcontainerbot/podman-next` copr. Correct me if I'm wrong.

I assume we have to override `trigger: pull_request` from the above `&copr` link.

I got the example from https://packit.dev/docs/configuration/#packit-service-jobs

lsm5:
> > @saschagrunert I'm cool with this but do we actually need the `trigger: commit` ? IIUC, it triggers builds after the commit is done, which is already being done by the webhook on `rhcontainerbot/podman-next` copr. Correct me if I'm wrong.
> 
> I assume we have to override `trigger: pull_request` from the above `&copr` link, so yes?! thinking
> 
> I got the example from https://packit.dev/docs/configuration/#packit-service-jobs

@saschagrunert i've pinged the packit people to understand `trigger: commit` cause right now I find it redundant along with the prior webhook. But if you wanna go ahead, by all means :slightly_smiling_face:  . If things blow up we'll soon find out :smile:

saschagrunert:
Sounds good, PTAL @haircommander @rphillips

lsm5:
so packit guys showered some knowledge. So, `trigger: commit` serves the same purpose as the webhook to `rhcontainerbot/podman-next` but with some additional status reporting on github itself. So, I think we can remove the webhook but we should also mention the copr project where the `after-commit` build should be sent to.

haircommander:
oops did I merge too early

lsm5:
na ... there can be another PR :D

lsm5:
only difference with `trigger: commit` would be `project: podman-next` and then we disable the earlier webhook

lsm5:
> only difference with `trigger: commit` would be `project: podman-next` and then we disable the earlier webhook

@saschagrunert webhook disabled. if you wanna do a new PR with the `project` change, please go ahead.

PR #1273, opened 2023-04-11
Fix get script blob verification

saschagrunert:
#### What type of PR is this?

/kind failing-test

Fixing the get-script CI. Taking the `GITHUB_ACTOR` into account is just not right, it should be always `containers/conmon-rs`.

None

None

```release-note
None
```


saschagrunert:

PR #1360, opened 2023-05-17
Check and fix typos in CI

saschagrunert:
#### What type of PR is this?

/kind ci

We now verify any typos in CI and fix them.

None
None

```release-note
Fixed API typos in golang client.
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1017,9 +1017,9 @@ func (c *ConmonClient) metadataBytes(ctx context.Context) ([]byte, error) {
 	return metadata, nil
 }
 
-// CreateaNamespacesConfig is the configuration for calling the
+// CreateNamespacesConfig is the configuration for calling the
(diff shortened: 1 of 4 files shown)


pkg/client/client.go
```diff
@@ -1017,9 +1017,9 @@ func (c *ConmonClient) metadataBytes(ctx context.Context) ([]byte, error) {
 	return metadata, nil
 }
 
-// CreateaNamespacesConfig is the configuration for calling the
+// CreateNamespacesConfig is the configuration for calling the
```
haircommander, line 1020: do we need type aliases to keep backward compatibility?

haircommander: though I don't know if we're using this anywhere so it may not actually matter..

saschagrunert: No right now it’s not being used.

PR #1440, opened 2023-06-27
Mention that the go client reuses a running server

mgjm:
#### What type of PR is this?

/kind documentation

Mention in the documentation that the go client will connect to an already running server instance.

When a server is already listening in the same `ServerRunDir`, the client reuses that server.

That also means that other options of the `ConmonServerConfig` will be ignored. The server will keep its settings from the first invocation.

```release-note
None
```


saschagrunert (approved):
Thank you!

/lgtm

PR #1497, opened 2023-07-11
Update makefile `update-proto`

mgjm:
#### What type of PR is this?

/kind other

This updates the `update-proto` rule in the Makefile to be more robust und work with go version 1.20.

- 880e0b7 creates an intermediate file in `internal/proto/conmon.capnp` with the required changes for go
  - Bug: If the `capnp compile` step failed the applied changes didn't get reverted, when tried again the patch was applied twice
  - Bug: The `git checkout` at the end did remove my uncommitted changes 😢
  - The new version no longer overwrites the existing file
  - The intermediate file gets removed by make automatically, even on error

- 17780b4 updates the make rule to be compatible with go version 1.20
  - `go install` no longer places source code in `$GOPATH/src`
  - The change uses the new `$GOMODCACHE = $GOPATH/pkg/mod` used by `go install` in module mode
  - And it uses the same go capnp version as specified in the `go.mod` file (instead of `latest`)

None

Verify that this change also works with older go versions (depending on the minimum required go version). I could not figure out in wich version go switched to the `GOMODCACHE` directory for `go install`.

And check that the grep rule works. Tested on Alpine and Arch Linux.

```release-note
None
```


saschagrunert (approved):
Thank you!

PR #1498, opened 2023-07-11
Use capnp map for tracing metadata

mgjm:
#### What type of PR is this?
/kind other

Currently (if tracing is enabled) the `metadata` field gets filled with a JSON encoded string map, that contains tracing metadata.

This PR changes this and uses capnp native maps and thereby eliminates the double encoding of JSON data inside of a capnp struct. Note that the [_proper way_](https://capnproto.org/language.html#generic-types) to implement maps in capnp is to use a `List` of string tuples.

Changes:
- Rename the existing `metadata` field to `metadataOld`
- Add a new `metadata` field of type `List[StringStringMapEntry]`
- Add utility functions to convert between a `List` of `MapEntry`s and rust types
- Remove `serde_json` from the server code
- Move capnp util function in go to `capnp_util.go` (slice and map encoding)
- Encode the metadata in the client in both formats (for backwards-compatibility)

Compatibility:
- I don't expect the conmon-rs server installed on a system to be newer than the client implementation.
- Eher the client and the server are installed with the same version (via os packages or installed from source)
- Or the client is newer (installed from source) and the server is older (installed via os packages)
- These cases are supported by the implemented backwards-compatibility in the client
- In the rare case that the server is on a newer version **only** tracing does not work (the server ignores the old field, everything else works as expected)

None

Are those compatibility requirements acceptable?

```release-note
None
```

### DIFF
--- a/conmon-rs/common/proto/conmon.capnp
+++ b/conmon-rs/common/proto/conmon.capnp
@@ -5,7 +5,8 @@ interface Conmon {
     # Version
     struct VersionRequest {
         verbose @0 :Bool;
-        metadata @1 :Data;
+        metadataOld @1 :Data; # deprecated
+        metadata @2 :Metadata; # Standard metadata to carry.
     }
 
     struct VersionResponse {
@@ -18,7 +19,7 @@ interface Conmon {
         rustVersion @6 :Text;
         cargoVersion @7 :Text;
         cargoTree @8 :Text;
-        metadata @9 :Data;
+        metadata @9 :Data; # deprecated (never used, remove in next major release)
     }
 
     version @0 (request: VersionRequest) -> (response: VersionResponse);
@@ -36,7 +37,8 @@ interface Conmon {
         cleanupCmd @7 :List(Text);
         globalArgs @8 :List(Text);
         commandArgs @9 :List(Text);
-        metadata @10 :Data;
+        metadataOld @10 :Data; # deprecated
+        metadata @11 :Metadata; # Standard metadata to carry.
     }
 
     struct LogDriver {
@@ -68,7 +70,8 @@ interface Conmon {
         timeoutSec @1 :UInt64;
         command @2 :List(Text);
         terminal @3 :Bool;
-        metadata @4 :Data;
+        metadataOld @4 :Data; # deprecated
+        metadata @5 :Metadata; # Standard metadata to carry.
     }
 
     struct ExecSyncContainerResponse {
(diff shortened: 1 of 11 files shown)


mgjm:
Third time's the charm. Fixed now **all** typos and lints.

conmon-rs/common/proto/conmon.capnp
```diff
@@ -5,7 +5,8 @@ interface Conmon {
     # Version
     struct VersionRequest {
         verbose @0 :Bool;
-        metadata @1 :Data;
+        metadataOld @1 :Data; # deprecated
```
haircommander, line 8: I believe it defeats the purpose of keeping the old metadata around if we change the name. Either we need to make a new name for the Metadata or just drop this value and claim incompatibilty with older clients.

mgjm: The name of a field is only relevant for the generated `proto` code. The binary encoding of capnp messages only depends on the field index `@1` (that didn't change).

Therefore the binary protocol stays compatible with older versions of the protocol. Only the library code that fills the generated `proto` structs with data needs to change, but that change is not user facing and therefore also not a breaking change.

haircommander: ah so if I have an old client and I'm accessing @1, I will just name it `metadata` and have no idea the server conceives of it as `metadataOld`?

haircommander:
@mgjm there's a merge conflict can you rebase please?

conmon-rs/common/proto/conmon.capnp
mgjm: Yes, that's how it is.

haircommander:
/approve

LGTM

@saschagrunert PTAL

saschagrunert (approved):

saschagrunert:
The linter is not happy:

```
error: this argument is a mutable reference, but not used mutably
   --> conmon-rs/server/src/child_reaper.rs:364:45
    |
364 |     async fn spawn_cleanup_process(raw_cmd: &mut Vec<String>) {
    |                                             ^^^^^^^^^^^^^^^^ help: consider changing to: `&Vec<String>`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
    = note: `-D clippy::needless-pass-by-ref-mut` implied by `-D warnings`
```

mgjm:
@saschagrunert Yeah, I haven't touched that code. Should I fix it anyway in this PR?

saschagrunert:
> @saschagrunert Yeah, I haven't touched that code. Should I fix it anyway in this PR?

Yes please, other PRs are affected as well by this.

mgjm:
Btw, this lint is a false positive:
```rs
    async fn spawn_cleanup_process(raw_cmd: &mut Vec<String>) {
        let mut cleanup_cmd = Command::new(raw_cmd.remove(0));
```

`raw_cmd.remove` is mutating function. But we can replace this with a slice access.

PR #1505, opened 2023-07-12
Fix forget_grandchild retain condition

mgjm:
#### What type of PR is this?
/kind bug

I think the retain condition in `ChildReaper::forget_grandchild` was flipped.

The function passed to retain specifies which elements to keep, but the specified condition returned which element we want to remove. ([docs](https://docs.rs/multimap/0.9.0/multimap/struct.MultiMap.html#method.retain))

That results in the removal of all `grandchildren` except the one we want to forget.

None

I assume that the `forget_grandchild` function should **remove** the grandchild with `pid == grandchild_pid` and not keep it.

There could be existing use cases that depend on the fact, that the ChildReaper forgot the wrong grandchildren by accident.

I'm not sure, but it could be the case that conmon-rs didn't terminate spawned childs on exit because of this and will do this now. This could be a braking change? If this is the case, we should mention it in the release note block below.

```release-note
None
```

### DIFF
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -157,8 +157,7 @@ impl ChildReaper {
         locked_grandchildren: &Arc<Mutex<MultiMap<String, ReapableChild>>>,
         grandchild_pid: u32,
     ) -> Result<()> {
-        let mut map = lock!(locked_grandchildren);
-        map.retain(|_, v| v.pid == grandchild_pid);
+        lock!(locked_grandchildren).retain(|_, v| v.pid != grandchild_pid);
         Ok(())
     }
 


conmon-rs/server/src/child_reaper.rs
```diff
@@ -158,7 +158,7 @@ impl ChildReaper {
         grandchild_pid: u32,
     ) -> Result<()> {
         let mut map = lock!(locked_grandchildren);
-        map.retain(|_, v| v.pid == grandchild_pid);
+        map.retain(|_, v| v.pid != grandchild_pid);
         Ok(())
```
saschagrunert, line 162: Hm, I think this is the right change, we can observe CRI-O CI after this merge to double check if it breaks anything.

Can we further simplify this please?

Something like this should be fine:
```rust
        Ok(lock!(locked_grandchildren).retain(|_, v| v.pid == grandchild_pid))
```

mgjm: I would prefer an explicit `Ok(())` statement, but I'm ok with both styles:
```rs
lock!(locked_grandchildren).retain(|_, v| v.pid != grandchild_pid);
Ok(())
```

What do you think?

saschagrunert: I'm not opposed to any of the above, thank you!

saschagrunert (approved):

PR #1553, opened 2023-08-01
Add support for file descriptor passing

mgjm:
#### What type of PR is this?

/kind feature

This PR adds support for file descriptors:
- New `StartFdSocket` RPC command
- `additionalFds` passed to the OCI runtime
- `leakFds` kept open by conmon as long as the container is running

I haven't found a good way to pass file descriptors via `SCM_RIGHTS` on the existing RPC connection. Therefore an additional socket is created for file descriptor passing. The protocol is documented in `conmon-rs/server/src/fd_socket.rs`.

RPC level workflow:
- The client calls `StartFdSocket` to start the fd socket server (if not already running) and gets the socket path
- The client send the file descriptors to the server via `SCM_RIGHTS` and gets slot numbers in return
- The client needs to keep the fd socket connection open until the fds are used
- The client calls `CreateContainer` with `additionalFds` or `leakFds`
- The client closes the fd socket connection (unused fds get closed at this point, the slot numbers are no longer valid)

GO client workflow:
- The user gets a `RemoteFDs` instance via the `client.RemoteFDs` method
- The user sends file descriptors via the `remoteFDs.Send` method and gets slot numbers in return
- The user calls `client.CreateContainer` with `AdditionalFDs` or `LeakFDs`
- The user calls `remoteFDs.Close` to close the connection

None

An alternative approach would be to hide the file descriptor logic from the public go client interface and just accept an file descriptor array in `CreateContainerConfig` and send the file descriptors in the `client.CreateContainer` method.

This would result in a cleaner API but the current design is simpler and allows reusing the `RemoteFDs` instance. But we create a new connection for each RPC request anyways, so does that performance improvement even matter?

```release-note
Add support for file descriptor passing
```

### DIFF
--- a/conmon-rs/common/proto/conmon.capnp
+++ b/conmon-rs/common/proto/conmon.capnp
@@ -41,6 +41,8 @@ interface Conmon {
         metadata @11 :Metadata; # Standard metadata to carry.
         envVars @12 :TextTextMap;
         cgroupManager @13 :CgroupManager;
+        additionalFds @14 :List(UInt64);
+        leakFds @15 :List(UInt64);
     }
 
     struct LogDriver {
@@ -168,6 +170,18 @@ interface Conmon {
     }
 
     createNamespaces @6 (request: CreateNamespacesRequest) -> (response: CreateNamespacesResponse);
+
+    ###############################################
+    # StartFdSocket
+    struct StartFdSocketRequest {
+        metadata @0 :Metadata; # Standard metadata to carry.
+    }
+
+    struct StartFdSocketResponse {
+        path @0 :Text; # The path to the fd socket.
+    }
+
+    startFdSocket @7 (request: StartFdSocketRequest) -> (response: StartFdSocketResponse);
 
     ###############################################
     # Helper types
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -4,7 +4,8 @@ use crate::{
     container_io::{ContainerIO, ContainerIOType, SharedContainerIO},
     oom_watcher::OOMWatcher,
 };
-use anyhow::{bail, format_err, Context, Result};
+use anyhow::{bail, Context, Result};
+use command_fds::{tokio::CommandFdAsyncExt, FdMapping};
 use getset::{CopyGetters, Getters, Setters};
 use libc::pid_t;
 use multimap::MultiMap;
@@ -19,6 +20,7 @@ use nix::{
 use std::{
     ffi::OsStr,
     fmt::Write,
+    os::fd::{AsRawFd, OwnedFd, RawFd},
     path::{Path, PathBuf},
     process::Stdio,
     str,
@@ -41,11 +43,8 @@ pub struct ChildReaper {
     grandchildren: Arc<Mutex<MultiMap<String, ReapableChild>>>,
 }
 
-macro_rules! lock {
-    ($x:expr) => {
-        $x.lock().map_err(|e| format_err!("{:#}", e))?
-    };
-}
+/// first usable file descriptor after stdin, stdout and stderr
+const FIRST_FD_AFTER_STDIO: RawFd = 3;
 
 impl ChildReaper {
     pub fn get(&self, id: &str) -> Result<ReapableChild> {
@@ -56,6 +55,7 @@ impl ChildReaper {
         Ok(r)
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub async fn create_child<P, I, S>(
         &self,
         cmd: P,
@@ -64,6 +64,7 @@ impl ChildReaper {
         container_io: &mut ContainerIO,
         pidfile: &Path,
         env_vars: Vec<(String, String)>,
+        additional_fds: Vec<OwnedFd>,
     ) -> Result<(u32, CancellationToken)>
     where
         P: AsRef<OsStr>,
@@ -81,8 +82,21 @@ impl ChildReaper {
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
             .envs(env_vars)
+            .fd_mappings(
+                additional_fds
+                    .iter()
+                    .enumerate()
+                    .map(|(i, fd)| FdMapping {
+                        parent_fd: fd.as_raw_fd(),
+                        child_fd: i as RawFd + FIRST_FD_AFTER_STDIO,
+                    })
+                    .collect(),
+            )?
             .spawn()
             .context("spawn child process: {}")?;
+
+        // close file descriptors after spawn
+        drop(additional_fds);
 
         let token = CancellationToken::new();
 
@@ -134,7 +148,11 @@ impl ChildReaper {
         Ok((grandchild_pid, token))
     }
 
-    pub fn watch_grandchild(&self, child: Child) -> Result<Receiver<ExitChannelData>> {
+    pub fn watch_grandchild(
+        &self,
+        child: Child,
+        leak_fds: Vec<OwnedFd>,
+    ) -> Result<Receiver<ExitChannelData>> {
         let locked_grandchildren = &self.grandchildren().clone();
         let mut map = lock!(locked_grandchildren);
         let mut reapable_grandchild = ReapableChild::from_child(&child);
@@ -148,6 +166,7 @@ impl ChildReaper {
         task::spawn(
             async move {
                 exit_tx.subscribe().recv().await?;
+                drop(leak_fds);
                 Self::forget_grandchild(&cleanup_grandchildren, pid)
             }
             .instrument(debug_span!("watch_grandchild", pid)),
--- a/conmon-rs/server/src/fd_socket.rs
+++ b/conmon-rs/server/src/fd_socket.rs
@@ -1,0 +1,261 @@
+//! File descriptor passing via UDS (unix socket).
+//!
+//! ## File descriptor socket protocol
+//!
+//! Connect via a unix `SOCK_SEQPACKET` socket.
+//!
+//! **Request:**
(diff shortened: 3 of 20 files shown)


internal/proto/conmon.capnp.go
```diff
@@ -4113,206 +4381,404 @@ func (p Conmon_createNamespaces_Results_Future) Response() Conmon_CreateNamespac
...
+	"H\x00\xde\xaa\xf8\xe6\xd2\xd3g\xa3S\xb6\x83\xfc\x03\xec" +
+	"c\xf4W\x84\xcb0\xa3+\x8aV\x01Z\xa3\x9f}\xed" +
+	"T\xd7\xb4\xf1;\xfcFO\x16\x9d&\xa3\x17\x99\xd1\xae" +
+	"\xbfS\xff\xe4\xd0\xfe\x9f\x11\x80\xf3\xb4\x01\xd6aq\x17" +
+	"J\xc3\x8bIUy\xf1m\x80\xd6\xaa\xc5\xc7\x9e]#" +
+	"_\xde\xd9\xcf\xf2\xaa\x8a\xbbQ\xba\xb5\x98\x96'\xcd\x9c" +
+	"\xb0\xefl]\xed\xae\xdc\xe5q\x84\x13\x097\x96\xe9\xbc" +
+	"\xa5\xf8Y@kr\xcf\x0b{\x1f\xfal\xf5\xdf\x13\x9a" +
+	"\xcb\xdd\xcc\x99\xe2e(}\\<\x14@\xba\xca\xd0\xbf" +
+	">ue\xdbC\xbfo\xdc\x93\xab\x9b'\xf4\xefJ\x0e" +
```
mgjm, line 4594: The "typos" GitHub action reports:
> "ue" should be "use" or "due"

... but this is a generated binary string, that just happens to contain the ASCII sequence "ue"

conmon-rs/common/proto/conmon.capnp
```diff
@@ -169,6 +171,18 @@ interface Conmon {
 
     createNamespaces @6 (request: CreateNamespacesRequest) -> (response: CreateNamespacesResponse);
 
+    ###############################################
+    # StartFdSocket
+    struct StartFdSocketRequest {
+        metadata @0 :Metadata; # Standard metadata to carry.
+    }
+
```
haircommander, line 179: should we just always start it when conmon-rs starts? that will prevent container creation requests from needing to check it's created before sending some down

conmon-rs/server/src/fd_socket.rs
```diff
@@ -0,0 +1,268 @@
...
+        } else {
+            *server = Server::start(path.clone(), self.clone()).await?.into();
+            Ok(path)
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn take(&self, slot: u64) -> Result<OwnedFd> {
+        lock!(self.state).take(slot)
+    }
```
haircommander, line 81: why the dead code?

haircommander:
a couple of notes, otherwise LGTM

conmon-rs/common/proto/conmon.capnp
mgjm: Some clients (e.g. CRI-O) do not need file descriptor passing at all or only for some containers (e.g. podman), therefore they also don't need the extra socket.

And when automatic shutdown is implemented, we can also stop the socket after a few seconds of inactivity.

conmon-rs/server/src/fd_socket.rs
mgjm: `take_all` was previously implemented as a loop over `take`, but refactored to reuse the same lock guard.

Now `take` is unused, but future use cases of the `FdSocket` could work with single file descriptors.

Therefore I think it would be nice to leave this method, so other people can use it in the future without having to understand the internals of the `FdSocket` struct.

haircommander:
@saschagrunert PTAL

conmon-rs/common/proto/conmon.capnp
haircommander: yeah I just think not starting it on startup makes some weird situations. does podman need to call StartFdSocket each time it runs to make sure? should the client keep track of it? how would the client restore that information? If we always start, we eliminate a lot of edge cases, for the cost of one unix socket running for a client that doesn't need it

mgjm: The current implementation stores no state in the client and always fetches the path from the server (and thereby starts the socket). This introduces an extra RPC call, but there should be no edge cases.

internal/proto/conmon.capnp.go
mgjm: Which option do we want?
1. Create a [`.typos.toml`](https://github.com/crate-ci/typos/blob/master/docs/reference.md) file (or `typos.toml` or `_typos.toml`) and specify `files.extend-exclude`.
2. Set the `files` option in the GitHub action to `. --exclude *.capnp.go`
3. Create a `.github/typos.toml` file and set the `config` option in the GitHub action to point to that file

conmon-rs/common/proto/conmon.capnp
saschagrunert: I see the benefits of both solutions and am not opinionated. :thinking:

conmon-rs/server/src/fd_socket.rs
```diff
@@ -0,0 +1,268 @@
...
+//!
+//! **Close file descriptors:**
+//! Received file descriptors will be closed on disconnect
+//! or with an empty request (`request_id` = 0 and `num_fds` = 0).
+//!
+//! An empty request does **not** receive a response.
+
+use crate::listener::{Listener, SeqpacketListener};
+use anyhow::format_err;
+use anyhow::Result;
```
saschagrunert, line 41: non blocking nit:
```suggestion
use anyhow::{format_err, Result};
```

conmon-rs/server/src/fd_socket.rs
```diff
@@ -0,0 +1,268 @@
...
+};
+use tokio::{runtime::Handle, sync::Mutex as AsyncMutex, task};
+use tokio_seqpacket::{ancillary::OwnedAncillaryMessage, UnixSeqpacket};
+use tracing::{debug_span, Instrument};
+
+macro_rules! lock {
+    ($x:expr) => {
+        $x.lock().map_err(|e| format_err!("{:#}", e))?
+    };
+}
```
saschagrunert, line 59: nit: Let's combine with the macro in `conmon-rs/server/src/child_reaper.rs`.

conmon-rs/server/src/fd_socket.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    fn take(&mut self, slot: u64) -> Result<OwnedFd> {
+        println!("take {slot}: {:?}", self.fds);
+        self.fds
+            .remove(&slot)
+            .ok_or_else(|| anyhow::anyhow!("no file descriptor in slot {slot}"))
+    }
+}
+
+#[derive(Debug, Default)]
```
saschagrunert, line 130: I don't think we need a default implementation for `Server`:
```suggestion
#[derive(Debug)]
```

conmon-rs/server/src/rpc.rs
```diff
@@ -236,6 +247,7 @@ impl conmon::Server for Server {
                         &mut container_io,
                         &pidfile,
                         env_vars,
+                        vec![],
```
saschagrunert, line 250: non blocking nit: I'd prefer having this an Option type `None`.

conmon-rs/common/proto/conmon.capnp
haircommander: let's leave as-is then

conmon-rs/server/src/rpc.rs
mgjm: May an ask why? I would add an Option if the `None` and `Some(vec![])` value would represent different states. But in this case both values would do the same. (and `vec![]` does not allocate either)

conmon-rs/server/src/child_reaper.rs
```diff
@@ -81,9 +85,22 @@ impl ChildReaper {
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
             .envs(env_vars)
+            .fd_mappings(
+                additional_fds
+                    .iter()
+                    .enumerate()
+                    .map(|(i, fd)| FdMapping {
+                        parent_fd: fd.as_raw_fd(),
+                        child_fd: (i + 3) as RawFd,
```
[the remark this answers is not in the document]
mgjm: It is the fist usable file descriptor after stdin, stdout and stderr. It's documented now.

saschagrunert (approved):

saschagrunert:
We probably should exclude the capnp file from the typos test: https://github.com/containers/conmon-rs/actions/runs/5815034612/job/15765845509?pr=1553

mgjm:
@saschagrunert Which [option to exclude capnp file from the typos test](https://github.com/containers/conmon-rs/pull/1553#discussion_r1287796434) do you prefer? I don't think the typos tool is used outside of CI, therefore I would lean towards option 3.

saschagrunert:
> @saschagrunert Which [option to exclude capnp file from the typos test](https://github.com/containers/conmon-rs/pull/1553#discussion_r1287796434) do you prefer? I don't think the typos tool is used outside of CI, therefore I would lean towards option 3.

Yeah I'm happy with the config file (option 3), too!

haircommander:
option 3 works for me!

saschagrunert (approved):
/lgtm

mgjm:
@saschagrunert the integration test finally succeeds. I needed to increase the MAX_RSS and the max allowed diff in the memory leak test

saschagrunert (approved):
I see, let's get that change in.

/lgtm

PR #1562, opened 2023-08-07
Fix clippy lint

saschagrunert:
#### What type of PR is this?

/kind failing-test

#### What this PR does / why we need it:
This fixes the redundant guard error:

```
error: redundant guard
   --> conmon-rs/server/src/container_io.rs:246:26
    |
246 |                 Ok(n) if n == 0 => {
    |                          ^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_guards
    = note: `-D clippy::redundant-guards` implied by `-D warnings`
help: try
    |
246 -                 Ok(n) if n == 0 => {
246 +                 Ok(0) => {
    |
```

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #1566, opened 2023-08-08
packit: Build PRs into default packit COPRs

martinpitt:
Building all PRs of all container projects into the same COPR does not properly isolate PRs from each other.

To avoid that, change the copr_build configuration to use the packit default COPRs, which are specific to the particular PR, and disappear after a few weeks. Depending projects should only run against what landed in conmon-rs/main i.e. the podman-next COPR.

#### What type of PR is this?

/kind failing-test

This is exactly the same as https://github.com/containers/crun/pull/1260 ; please see the discussion there, this change needs to be applied to all container projects, and then all land together. Let's keep all the relevant discussions and tracking in  the crun PR please.

None

Discussed with @lsm5 in https://github.com/containers/crun/pull/1260

```release-note
None
```

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,46 +2,41 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
-# Pull request build targets can be found at:
-# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
-
 # Main branch commit build targets can be found at:
 # https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/
 
 specfile_path: conmon-rs.spec
 
-jobs:
-  - &copr
-    job: copr_build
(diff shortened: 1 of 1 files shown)


martinpitt:
[one integration test failure](https://github.com/containers/conmon-rs/actions/runs/5795995554/job/15708788883?pr=1566): 

> [FAIL] ConmonClient [JustAfterEach] ExecSync Stress should handle many requests with terminal

I'll retry this once to compare, but waiting for the other builds first.

martinpitt:
Clippy is fixed in #1562 , so this seems okay now. The centos-stream+epel-next-8 build is still going on (dnf install is ludicrously slow, networking problems?)

lsm5 (approved):
/lgtm

Thanks!

saschagrunert (approved):
Cool, thank you!

PR #1590, opened 2023-08-15
Add automatic shutdown after inactivity

mgjm:
#### What type of PR is this?

/kind feature

This adds a new option to enable automatic shutdown: After a configurable period of inactivity the server automatically shuts itself down.

The following activity is tracked and prevents shutdown:
- A connected RPC client
- A running `ReapableChild` of the child reaper (= a running container)
- A running cleanup process

The new FD socket server also shuts itself down after 3 seconds of inactivity (= no connected clients).

None

The FD socket inactivity period is hardcoded as 3 seconds. This is enough time for the client to connect to the FD socket path, after the client invoked the `StartFdSocket` RPC method. This could also be configurable.

```release-note
Add an option to enable automatic shutdown after a configurable period of inactivity
```

### DIFF
--- a/conmon-rs/server/src/config.rs
+++ b/conmon-rs/server/src/config.rs
@@ -12,7 +12,7 @@ macro_rules! prefix {
     };
 }
 
-#[derive(CopyGetters, Debug, Deserialize, Eq, Getters, Parser, PartialEq, Serialize, Setters)]
+#[derive(CopyGetters, Debug, Deserialize, Getters, Parser, PartialEq, Serialize, Setters)]
 #[serde(rename_all = "kebab-case")]
 #[command(
     after_help("More info at: https://github.com/containers/conmon-rs"),
@@ -125,6 +125,16 @@ pub struct Config {
     )]
     /// OpenTelemetry GRPC endpoint to be used for tracing.
     tracing_endpoint: String,
+
+    #[get_copy = "pub"]
+    #[arg(
+        default_value_t,
+        env(concat!(prefix!(), "SHUTDOWN_DELAY")),
+        long("shutdown-delay"),
+        value_name("DELAY"),
+    )]
+    /// Automatically stop conmon-rs after DELAY seconds of inactivity. (0 = disabled)
+    shutdown_delay: f64,
 }
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Subcommand)]
--- a/conmon-rs/server/src/inactivity.rs
+++ b/conmon-rs/server/src/inactivity.rs
@@ -1,0 +1,128 @@
+use std::{
+    future, process,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use tokio::sync::{futures::Notified, Notify};
+
+/// Track activity and reacto to inactivity.
+///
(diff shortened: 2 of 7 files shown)


conmon-rs/server/src/inactivity.rs
```diff
@@ -0,0 +1,128 @@
...
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use tokio::sync::{futures::Notified, Notify};
+
+/// Track activity and reacto to inactivity.
```
saschagrunert, line 12:
```suggestion
/// Track activity and react to inactivity.
```

conmon-rs/server/src/config.rs
```diff
@@ -125,6 +125,16 @@ pub struct Config {
...
+
+    #[get_copy = "pub"]
+    #[arg(
+        default_value_t,
+        env(concat!(prefix!(), "SHUTDOWN_DELAY")),
+        long("shutdown-delay"),
+        value_name("DELAY"),
+    )]
+    /// Automatically stop conmon-rs after DELAY seconds of inactivity. (0 = disabled)
+    shutdown_delay: f64,
```
saschagrunert, line 137: I think we should make this something like [Option<`humantime::Duration`>](https://docs.rs/humantime/latest/humantime/struct.Duration.html)

pkg/client/client.go
```diff
@@ -95,6 +95,11 @@ type ConmonServerConfig struct {
 
 	// Tracing can be used to enable OpenTelemetry tracing.
 	Tracing *Tracing
+
+	// ShutdownDelay an be used to automatically stop the conmonrs server after DELAY seconds of inactivity.
```
saschagrunert, line 99: Can we document what _inactivity_ means? Like, which actions are considering conmon-rs to be active?

conmon-rs/server/src/config.rs
mgjm: Hmm... the string representation of a go `time.Duration` is not compatible with the `humantime::Duration` parser. But we could write a custom to string method in go. Do we expect the conmon server to be invoked by humans or only by the go client?

haircommander: IMO we should only support go client

mgjm: And even for humans it is still relatively ergonomic to specify the duration in (fractional) seconds.  
`1.5` vs `1s 500ms`  
Especially considering that I expect the most common durations to be in the range of a few full seconds.

pkg/client/client.go
rphillips: +1 to what inactivity means. 

What use cases is this feature trying to solve?

conmon-rs/server/src/config.rs
saschagrunert: Is specifying milliseconds really a use case? I'm also fine with using seconds only.

saschagrunert:
> Is there a reason we open a new connection for each RPC call? If not, the solution would be open a connection in the `ConmonClient` constructor and always use that connection.

I think the main intention was for simplicity. If the connection dies then the client would have to create a new one.

saschagrunert:
Closing, please reopen and rebase if you'd like to still work on that.

PR #1595, opened 2023-08-15
Change resize channel to receive only

mgjm:
#### What type of PR is this?
/kind cleanup

The `Resize` channel in the `AttachConfig` go struct was not restricted. This PR changes the channel to a receive only channel.

```diff
-	Resize chan resize.TerminalSize
+	Resize <-chan resize.TerminalSize
```

The channel is only used for receive operations.

**Note:** This is technically a breaking change as the following code would no longer work:

```go
config := new(client.AttachConfig)
config.Resize = make(chan resize.TerminalSize, 1)
config.Resize <- size
```

But this is an uncommon usage pattern. The following pattern is more expected (used by CRI-O) and is unaffected by this change:
```go
resize := make(chan resize.TerminalSize, 1)
resize <- size

config := new(client.AttachConfig)
config.Resize = resize
```

None

```release-note
None
```

### DIFF
--- a/pkg/client/attach.go
+++ b/pkg/client/attach.go
@@ -74,7 +74,7 @@ type AttachConfig struct {
 	Passthrough bool
 
 	// Channel of resize events.
-	Resize chan resize.TerminalSize
+	Resize <-chan resize.TerminalSize
 
 	// The standard streams for this attach session.
 	Streams AttachStreams


saschagrunert (approved):
LGTM, PTAL @haircommander @rphillips because of the breaking change

PR #1596, opened 2023-08-16
Update golangci-lint to v1.54.1

saschagrunert:
#### What type of PR is this?

/kind cleanup

This applies the reported lints as well as updated the config.

None
None

```release-note
None
```


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containers/conmon-rs
 
-go 1.18
+go 1.20
```
saschagrunert, line 3: required by `errors.Join`, but we already use go 1.20 in CI so I assume this is fine :shrug:

PR #1654, opened 2023-08-29
Fix stop after EOF error condition

mgjm:
#### What type of PR is this?
/kind bug

While implementing attach for podman the attach stream would never terminate (at least without setting `cfg.StopAfterStdinEOF`, but then it would not wait for all output).

This change fixes this issue in my case, but it might be the wrong fix. Maybe it should instead be just the following:
```go
if er == io.EOF {
    return nil
}
```

None

```release-note
None
```

### DIFF
--- a/pkg/client/attach.go
+++ b/pkg/client/attach.go
@@ -270,7 +270,7 @@ func (c *ConmonClient) redirectResponseToOutputStreams(
 				return nil
 			}
 		}
-		if er == io.EOF || (cfg.ContainerStdin && !cfg.StopAfterStdinEOF) {
+		if er == io.EOF && cfg.ContainerStdin && !cfg.StopAfterStdinEOF {
 			return nil
 		}
 		if errors.Is(er, syscall.ECONNRESET) {


pkg/client/attach.go
```diff
@@ -270,7 +270,7 @@ func (c *ConmonClient) redirectResponseToOutputStreams(
 				return nil
 			}
 		}
-		if er == io.EOF || (cfg.ContainerStdin && !cfg.StopAfterStdinEOF) {
+		if er == io.EOF && cfg.ContainerStdin && !cfg.StopAfterStdinEOF {
```
saschagrunert, line 273: Hm, we have to test if this works with CRI-O, I know we had some issues with that in the past.

mgjm: The more I think about it, it probably should just be
```go
if er == io.EOF {
    return nil
}
```

saschagrunert: `cfg.ContainerStdin && !cfg.StopAfterStdinEOF` feels like a hack, is that related to https://github.com/containers/conmon-rs/pull/1655?

rphillips: Likely we need to detect the EOF and then drain the streams.

saschagrunert:
@mgjm are you still working on this?

saschagrunert:
Closing, please reopen and rebase if you'd like to still work on that.

PR #1655, opened 2023-08-29
Fix early stdio close due to multiple ownership

mgjm:
#### What type of PR is this?
/kind bug

The `ContainerIO::read_loop_stdin` function was taking the fd to stdin as a `RawFd` and took ownership of the file descriptor. This worked for piped IO (`streams.rs`) but not for tty based IO (`terminal.rs`).

With tty based IO the same file descriptor was used as an `AsyncFd` and passed to `read_loop_stdin` this resulted in two owners of the same file descriptor. The first owner (stdin in most cases) would close the fd on drop. The second owner would stall as it won't be notified that the fd closed and once the second owner would drop it would try to close the same fd again. Which could have been reused in the meantime.

This change refactors `ContainerIO::read_loop_stdin` to take an implementation of `AsyncWrite` and introduces a new `TerminalFd` which wraps the pty file descriptor and implements  `AsyncRead` and `AsyncWrite` for references to itself. This allows us to use reference counting to close the file descriptor only after both the `read_loop_stdin` and `read_loop` terminate.

None

```release-note
None
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -365,7 +365,6 @@ dependencies = [
  "time",
  "tokio",
  "tokio-eventfd",
- "tokio-fd",
  "tokio-seqpacket",
  "tokio-util",
  "tracing",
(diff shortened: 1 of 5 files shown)


saschagrunert (approved):
That's a good one, thank you for this fix!

haircommander:
/lgtm

great find!

PR #1689, opened 2023-09-12
Fix get script architecture

saschagrunert:
#### What type of PR is this?

/kind cleanup

We should print the right environment variable for the architecture and also use the right fallback.

None
None

```release-note
None
```


saschagrunert:

PR #2078, opened 2024-03-21
make ginkgo version keep up to the version in go.mod

bitoku:
#### What type of PR is this?

/kind dependency-change

ginkgo version in make file wasn't matched with the version in go.mod.
It seems ginkgo version update is fast so it would be good to keep the version up to date by running `go list`.

None

```release-note
None
```


saschagrunert (approved):
Thanks!

PR #2136, opened 2024-04-22
Switch to go 1.22

saschagrunert:
#### What type of PR is this?

/kind ci

Fix to the latest go version and adjust the failing test.

None
None

```release-note
None
```

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -695,7 +695,11 @@ var _ = Describe("JSONLogger", func() {
 		for _, terminal := range []bool{true, false} {
 			It(testName("should log in JSON format", terminal), func() {
 				tr = newTestRunner()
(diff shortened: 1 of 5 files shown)


saschagrunert:

pkg/client/client_test.go
```diff
@@ -707,7 +711,8 @@ var _ = Describe("JSONLogger", func() {
 						Path: tr.logPath(),
 					}},
 				})
-				Expect(err).To(HaveOccurred())
+				Expect(err).NotTo(HaveOccurred())
+				tr.startContainer(sut)
```
haircommander, line 715: it's not clear to me why this switched

saschagrunert: I changed the container above to actually be valid. Fun fact is that the for loop `for _, terminal := range []bool{true, false} {` in go 1.21 (ginkgo related?) does not work for `true`, but it works in go 1.22.

Issue #2217, opened 2024-05-24
Failed to start the crio service with error meaasge: " parse conmon version: No Major.Minor.Patch elements found"


Bevisy:
### What happened?

1

### What did you expect to happen?

1

### How can we reproduce it (as minimally and precisely as possible)?

1

### conmon-rs version

<details>

```console
$ conmonrs --version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Additional environment details (AWS, VirtualBox, physical, etc.)

<details>

</details>

PR #2218, opened 2024-05-24
fix: add additional string to meet crio's version parsing requirements

Bevisy:
add additional string to meet crio's version parsing requirements

refer to https://github.com/cri-o/cri-o/issues/8214#issuecomment-2129203924

Fixes: https://github.com/cri-o/cri-o/issues/8214, https://github.com/containers/conmon-rs/issues/2217

### DIFF
--- a/conmon-rs/server/src/version.rs
+++ b/conmon-rs/server/src/version.rs
@@ -58,7 +58,7 @@ impl Version {
 
     /// Print the version information to stdout.
     pub fn print(&self) {
-        println!("version: {}", self.version());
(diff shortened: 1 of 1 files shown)


Bevisy:
> I'm not opposed to this, but fixing this will kick the can down the road, so to speak, as the real issue was the wrong runtime_type. Do you think we still want this @Bevisy ?

I agree with you. This approach indeed conceals the real issue. We should add document or error log in cri-o to clearly indicate that the problem lies in the mismatch configuration between `runtime_type` and `conmonrs`.

PR #2317, opened 2024-07-21
Update go-capnp package to fix build issue on Go 1.23

kwilczynski:
#### What type of PR is this?

/kind dependency-change

When building the [go-capnp](https://github.com/capnproto/go-capnp) using Go 1.23 or Go tip, the following build error will be reported (an example build failure from the CRI-O project):

```console
# capnproto.org/go/capnp/v3
vendor/capnproto.org/go/capnp/v3/list.go:1087:10: invalid composite literal type T
```

Thus, upgrade the go-capnp version to [v3.0.1-alpha.2](https://github.com/capnproto/go-capnp/releases/tag/v3.0.1-alpha.2), which would fix the build issue on a newer version of Go.

Related:

- https://github.com/capnproto/go-capnp/pull/576
- https://github.com/cri-o/cri-o/issues/8373

None

None

```release-note
None
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,7 +321,7 @@ jobs:
       - name: create symlink
         run: sudo ln -f -s $(go env GOROOT)/bin/* /usr/bin/
       - name: Integration tests
-        run: RUNTIME_PATH="/usr/sbin/runc" make integration-static
+        run: sudo -E RUNTIME_PATH="/usr/sbin/runc" make integration-static
       - name: Chown cache
         run: |
           sudo chown -R $(id -u):$(id -g) ~/go/pkg/mod
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Select Toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: ${{ env['ACTION_MSRV_TOOLCHAIN']  }}
+          toolchain: ${{ env['ACTION_MSRV_TOOLCHAIN'] }}
           default: true
           override: true
           components: rustfmt
@@ -86,7 +86,7 @@ jobs:
           # Ubuntu 22.04 glibc static is not compatible with rustc 1.58.1 (see
           # ACTION_MSRV_TOOLCHAIN). Means we now just use the latest one, since
           # the static builds are made for the community.
-          toolchain: stable
+          toolchain: ${{ env['ACTION_MSRV_TOOLCHAIN'] }}
           default: true
           override: true
           components: rustfmt
@@ -207,7 +207,7 @@ jobs:
       - name: Select Toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: ${{ env['ACTION_MSRV_TOOLCHAIN']  }}
+          toolchain: ${{ env['ACTION_MSRV_TOOLCHAIN'] }}
           default: true
           override: true
           components: rustfmt
@@ -231,13 +231,13 @@ jobs:
       - name: Select Toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: ${{ env['ACTION_MSRV_NIGHTLY_TOOLCHAIN']  }}
+          toolchain: ${{ env['ACTION_MSRV_NIGHTLY_TOOLCHAIN'] }}
           default: true
           override: true
           components: clippy, rustfmt
       - name: Clippy Lint
         run: |
-          cargo +${{ env['ACTION_MSRV_NIGHTLY_TOOLCHAIN']  }} clippy \
+          cargo +${{ env['ACTION_MSRV_NIGHTLY_TOOLCHAIN'] }} clippy \
             --all-targets --all-features -- -D warnings
 
   lint-rustfmt:
@@ -248,7 +248,7 @@ jobs:
       - name: Select Toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: ${{ env['ACTION_MSRV_TOOLCHAIN']  }}
+          toolchain: ${{ env['ACTION_MSRV_TOOLCHAIN'] }}
           default: true
           override: true
           components: rustfmt
@@ -264,7 +264,7 @@ jobs:
       - name: Select Toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: ${{ env['ACTION_MSRV_NIGHTLY_TOOLCHAIN']  }}
+          toolchain: ${{ env['ACTION_MSRV_NIGHTLY_TOOLCHAIN'] }}
           override: true
       - name: Install rustfmt
         shell: bash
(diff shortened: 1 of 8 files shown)


saschagrunert:
Integration and critest seems to fail, do you know why?

kwilczynski:
> Integration tests seems to fail, do you know why?

@saschagrunert, I will look into it.

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   GO_VERSION: '1.22'
-  ACTION_MSRV_TOOLCHAIN: 1.66.1
+  ACTION_MSRV_TOOLCHAIN: 1.67.0
```
saschagrunert, line 12: We need to clarify with downstream if we can do that. I doubt looking at the latest conmon-rs builds in brew.

kwilczynski: No worries. I moved my local environment to v1.66.1, which resolves the time crate build issue. too.

kwilczynski: For posterity, the following was the time crate build issue:

Screenshot from 2024-07-30 16-54-37

kwilczynski: Actually, I might have spoken too soon. We seem to have the same issue too, per:

https://github.com/containers/conmon-rs/actions/runs/10164569684/job/28110266502?pr=2317#step:6:139

kwilczynski: I don't seem to have this issue locally any more.

kwilczynski: OK. We might need to pin down Rust version for the static build job too, per:

```console
rustc 1.79.0 (129f3b996 2024-06-10)
/home/runner/.cargo/bin/rustup -V
rustup 1.27.1 (54dd3d00f [20](https://github.com/containers/conmon-rs/actions/runs/10164569684/job/28110266502?pr=2317#step:5:21)24-04-24)
```

For reference:

- https://github.com/containers/conmon-rs/actions/runs/10164569684/job/28110266502?pr=2317#step:5:20

kwilczynski: Or maybe I broke it when I removed sudo... let me check.

kwilczynski: No. Wasn't me. :sweat_smile:

kwilczynski: @saschagrunert, the RPM builds on a more modern Fedora version have the same issue.

How do you want to go about it? Since time crate upgrade is a no-no? (requires slightly newer Rust)

saschagrunert: I was thinking about this yesterday. We could either fork the crate and patch it, see: https://github.com/saschagrunert/time/tree/v0.3.23-fix

But Cargo is a bit picky on versions, so we would have to replace it for all dependencies, not just the direct ones.

saschagrunert: Probably like this: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#using-patch-with-multiple-versions

saschagrunert:
Yay, green CI! :green_circle:

Cargo.toml
```diff
@@ -12,3 +12,6 @@ lto = true
 opt-level = "z"
 incremental = true
 codegen-units = 1
+
+[patch.crates-io]
```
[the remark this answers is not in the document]
kwilczynski: Done.

saschagrunert:
@kwilczynski is this still WIP?

pkg/client/client.go
```diff
@@ -887,8 +887,10 @@ func (c *ConmonClient) ExecSyncContainer(ctx context.Context, cfg *ExecSyncConfi
 
 	execContainerResult := &ExecContainerResult{
 		ExitCode: resp.ExitCode(),
-		Stdout:   stdout,
-		Stderr:   stderr,
+		// Copy content of both slices as the `capnp.ReleaseFunc` will
+		// zero them when the deferred call to `free()` takes place.
+		Stdout:   append(stdout[:0:0], stdout...),
+		Stderr:   append(stderr[:0:0], stderr...),
```
kwilczynski, line 893: **To the reviewer:**

Since the changes from the following Pull Request:

- https://github.com/capnproto/go-capnp/pull/489

Following these changes referenced above, the `ReleaseFunc` would clean (or reset, if you wish) the data the promise returns. This is a problem as slices are passed using a reference, rather than copied like the primitive types. As such, the slice which holds the data will be reset to its default state behind the scenes once the `free()` callback from the affected code is invoked.

Thus, we have two options:

1. Copy the content of the stderr and stdout slices
1. Change the API such that the function would return the `ReleaseFunc` callback too

The #1 is simple and keeps the implementation details still hidden, but would waste a little bit of memory in the process, whereas the #2 would require calls to not forget to call the `ReleaseFunc` once the results are no longer needed.

kwilczynski:
> @kwilczynski is this still WIP?

@saschagrunert, should be ready for review, etc.

saschagrunert (approved):
Thanks!

pkg/client/client.go
saschagrunert: Let's go with option 1 for now.

kwilczynski:
@saschagrunert, do you think we could cat a new release?

saschagrunert:
> @saschagrunert, do you think we could cut a new release?

Yes, let me first check if we can / have to fix https://github.com/cri-o/cri-o/issues/8411. I'm not even sure if that's a conmon-rs issue.

kwilczynski:
> > @saschagrunert, do you think we could cut a new release?
> 
> Yes, let me first check if we can / have to fix [cri-o/cri-o#8411](https://github.com/cri-o/cri-o/issues/8411). I'm not even sure if that's a conmon-rs issue.

@saschagrunert, the exit code and time out boolean value should be copied so the correct exit code should be reflected. We have a check in a test to verify this, per:

- https://github.com/containers/conmon-rs/blob/main/pkg/client/client_test.go#L389
- https://github.com/containers/conmon-rs/blob/main/pkg/client/client_test.go#L420

I also doubt that our change here resolved anything for the other issue.

buckaroogeek:
Would it be possible to retrofit this PR to CRI-O v1.29? For a variety of reasons Fedora 41 provides Kubernetes 1.29  as the default `kubernetes` rpm and therefore needs to also provide CRI-O v1.29 in the default repositories. Fedora 41 also uses go 1.23 as the golang version. As a result, CRI_O v1.29 will not build as it still has the older version of go-capnp.

This request is not urgent as, starting with Fedora 41, I am providing versioned RPMS for Kubernetes and CRI-O (and CRI-Tools) so that all 3 supported versions of Kubernetes, CRI-O, and CRI-Tools will be available concurrently. For example there are now `cri-o1.29` and `cri-1.30` rpms in F41 (and F42) and I am prepared to push `cri-o1.31` as soon as it becomes available.  

However, the legacy default `crio` package cannot be updated due to the go 1.23 and go-capnp problem.  As noted this is not urgent as users have the versioned rpms as an alternative in addition to the rpms the CRI-O team provides.

thanks!

PR #2353, opened 2024-08-01
Fix OOM slice removal race

saschagrunert:
#### What type of PR is this?

/kind bug

If the slice is already removed then we mostly encounter two different errors:

- `get next line: No such device (os error 19)`
- `open memory events file: /sys/fs/cgroup/test.slice/crio-$ID.scope/memory.events: No such file or directory (os error 2)`

To avoid such a race we now check after the errors if the file still exists. If not, then we assume an OOM.

Fixes https://github.com/cri-o/cri-o/issues/8411

None

```release-note
Fixed OOM slice removal race to not report an out of memory event.
```

### DIFF
--- a/conmon-rs/server/src/oom_watcher.rs
+++ b/conmon-rs/server/src/oom_watcher.rs
@@ -224,13 +224,35 @@ impl OOMWatcher {
                     if let Err(e) = tx.try_send(OOMEvent{ oom: false }) {
                         error!("try_send failed: {:#}", e);
                     };
+
+                    debug!("Last resort check for OOM");
+                    match Self::check_for_oom(&memory_events_file_path, last_counter).await {
+                        Ok((counter, is_oom)) => {
+                            if !is_oom {
+                                debug!("No OOM found in event");
+                                break;
+                            }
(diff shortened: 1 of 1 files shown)


saschagrunert:

conmon-rs/server/src/oom_watcher.rs
```diff
@@ -252,8 +253,19 @@ impl OOMWatcher {
                                     };
                                 }
                                 Err(e) => {
-                                    error!("Checking for OOM failed: {}", e);
-                                    match tx.try_send(OOMEvent{ oom: false }) {
+                                    error!("Checking for OOM failed: {:#}", e);
+
+                                    let mut oom = false;
+                                    if !memory_events_file_path.exists() {
```
rphillips, line 259: I'm not sure which exists() this is... A better call might be tokio's async call (untested):

```suggestion
 !tokio::fs::metadata("/some/file/path.txt").await?.is_ok()
```

saschagrunert: Yep, done :+1:

rphillips:
Nice!
/lgtm

PR #2415, opened 2024-09-10
Run waitpid on childs to ensure no zombies are left behind

saschagrunert:
#### What type of PR is this?

/kind bug

We can call waitpid on the process to reparent and ensure no zombies are left behind.

Fixes https://github.com/containers/crun/issues/1482
Fixes https://github.com/containers/conmon-rs/issues/2242

None

```release-note
Fixed issue that containers cannot be removed if init command was not found.
```

### DIFF
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -13,7 +13,7 @@ use nix::{
     errno::Errno,
     sys::{
         signal::{kill, Signal},
-        wait::{waitpid, WaitStatus},
+        wait::{waitpid, WaitPidFlag, WaitStatus},
     },
     unistd::{getpgid, Pid},
 };
@@ -137,6 +137,9 @@ impl ChildReaper {
             // meaning there is no other entity that could cancel the read_loops.
             token.cancel();
 
+            // Wait to ensure that all children do not become zombies.
+            Self::check_child_processes();
+
             bail!(err_str)
         }
 
@@ -147,6 +150,34 @@ impl ChildReaper {
             .context(format!("grandchild pid parse error {}", pidfile.display()))?;
 
         Ok((grandchild_pid, token))
+    }
+
+    fn check_child_processes() {
+        debug!("Checking child processes");
+        let pid = Pid::from_raw(-1);
+        loop {
+            match waitpid(pid, WaitPidFlag::WNOHANG.into()) {
+                Ok(WaitStatus::Exited(p, code)) => {
+                    debug!("PID {p} exited with status: {code}");
+                    break;
+                }
+                Ok(WaitStatus::StillAlive) => {
+                    debug!("PID {pid} is still in same state");
+                    break;
+                }
+                Ok(_) => {
+                    continue;
+                }
+                Err(Errno::EINTR) => {
+                    debug!("Retrying on EINTR for PID {pid}");
+                    continue;
+                }
+                Err(err) => {
+                    error!("Unable to waitpid on {:#}", err);
+                    break;
+                }
+            };
+        }
     }
 
     pub fn watch_grandchild(


saschagrunert:
Fixes https://github.com/containers/conmon-rs/issues/2242

conmon-rs/server/src/child_reaper.rs
```diff
@@ -137,6 +137,9 @@ impl ChildReaper {
             // meaning there is no other entity that could cancel the read_loops.
             token.cancel();
 
+            // Ensure to wait for all childs to not become zombies.
```
kwilczynski, line 140:
```suggestion
            // Wait to ensure that children do not become zombies.
```

conmon-rs/server/src/child_reaper.rs
```diff
@@ -149,6 +152,34 @@ impl ChildReaper {
...
+        debug!("Checking child processes");
+        let pid = Pid::from_raw(-1);
+        loop {
+            match waitpid(pid, WaitPidFlag::WNOHANG.into()) {
+                Ok(WaitStatus::Exited(p, code)) => {
+                    debug!("PID {p} exited with status: {code}");
+                    break;
+                }
+                Ok(WaitStatus::StillAlive) => {
+                    debug!("Process is still in same state, breaking");
```
kwilczynski, line 165:
```suggestion
                    debug!("PID {p} is still in same state");
```

conmon-rs/server/src/child_reaper.rs
```diff
@@ -149,6 +152,34 @@ impl ChildReaper {
...
+                    break;
+                }
+                Ok(_) => {
+                    continue;
+                }
+                Err(Errno::EINTR) => {
+                    debug!("Failed to wait for PID on EINTR, retrying");
```
kwilczynski, line 172:
```suggestion
                    debug!("Retrying on EINTR for PID {pid}");
```

saschagrunert: Fixed as suggested :+1:

conmon-rs/server/src/child_reaper.rs
saschagrunert: Fixed as suggested :+1:

conmon-rs/server/src/child_reaper.rs
saschagrunert: Fixed as suggested :+1:

conmon-rs/server/src/child_reaper.rs
```diff
@@ -149,6 +152,34 @@ impl ChildReaper {
...
+    fn check_child_processes() {
+        debug!("Checking child processes");
+        let pid = Pid::from_raw(-1);
+        loop {
+            match waitpid(pid, WaitPidFlag::WNOHANG.into()) {
+                Ok(WaitStatus::Exited(p, code)) => {
+                    debug!("PID {p} exited with status: {code}");
+                    break;
+                }
+                Ok(WaitStatus::StillAlive) => {
```
kwilczynski, line 164: Do we also want to match here on  `WaitStatus::Stopped()` and `WaitStatus::Continued()`? I am thinking about a case when the process suddenly _goes away_ for reasons outside of conmon-rs' influence.

kwilczynski: On the other hand, perhaps leaving it `Ok()` is fine, to skip over, so to speak.

kwilczynski: Basically, I am not sure if `WaitStatus::StillAlive` covers stopped and continued process, and if not, then whether we should break here for such or continue waiting? @saschagrunert, thoughts?

kwilczynski: Makes sense. Thank you!

Issue #2636, opened 2025-07-30
Add scalability tests to CI


saschagrunert:
Goal is to add a series of tests which test the scalability of conmon-rs.

saschagrunert:
Good, conmon-rs scales linearly with the amount of added containers:

<img width="1499" height="1189" alt="Image" src="https://github.com/user-attachments/assets/ddea44f6-267a-4b68-bbe1-91f42f6aa268" />

PR #2640, opened 2025-08-01
Add scalability tests

saschagrunert:
#### What type of PR is this?

/kind ci

Adding a first set of scalability tests and its CI infra.

Part of #2636 
None

```release-note
None
```

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -45,7 +45,7 @@ var _ = Describe("ConmonClient", func() {
 	})
 
 	AfterEach(func() {
-		Expect(tr.rr.RunCommand("delete", "-f", tr.ctrID)).To(Succeed())
+		Expect(tr.rr.RunCommand("delete", "-f", tr.ctrID())).To(Succeed())
 		if sut != nil {
 			Expect(sut.Shutdown()).To(Succeed())
 		}
@@ -109,7 +109,7 @@ var _ = Describe("ConmonClient", func() {
 				tr.createRuntimeConfig(terminal)
 				sut = tr.configGivenEnv()
 				tr.createContainer(sut, terminal)
-				tr.startContainer(sut)
+				tr.startContainer(tr.ctrID())
 				Expect(fileContents(tr.exitPath())).To(Equal("0"))
 			})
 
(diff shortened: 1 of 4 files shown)


saschagrunert:
/hold
needs https://github.com/containers/conmon-rs/pull/2641

pkg/client/client_test.go
```diff
@@ -903,3 +904,43 @@ var _ = Describe("StreamingServer", func() {
...
+		evalVMRss(maxRSSKB + ctrCount*40)
+	},
+		Entry("without terminal and 10 containers", false, 10),
+		Entry("with terminal and 10 containers", true, 10),
+		Entry("without terminal and 50 containers", false, 50),
+		Entry("with terminal and 50 containers", true, 50),
+		Entry("without terminal and 100 containers", false, 100),
+		Entry("with terminal and 100 containers", true, 100),
+	)
+})
```
saschagrunert, line 946: The actual test cases.

saschagrunert:
@haircommander PTAL 

7a23b17de2db58d12f3561351c793e05

haircommander:
I think this is good to have the numbers spelled out like this, but it's not exactly what I was thinking. specifically, I'm worried about us switching to conmon-rs and then suddenly using way more memory on the node and needing to reserve more for system reserved. I am not sure these tests should run here, and I feel bad about asking after you've done all this work, but I'm wondering if we can have a comparison suite between conmon and conmon-rs to make sure total memory usage in the system slice doesn't increase a wild amount. maybe cri-o would be a better place for it?

also: this test is reported as flakey `↺ [FLAKEY TEST - TOOK 3 ATTEMPTS TO PASS] [65.455 seconds]` so that's a risk to what currently exists

sorry to be pedantic!

/cat

saschagrunert:
> I think this is good to have the numbers spelled out like this, but it's not exactly what I was thinking. specifically, I'm worried about us switching to conmon-rs and then suddenly using way more memory on the node and needing to reserve more for system reserved. I am not sure these tests should run here, and I feel bad about asking after you've done all this work, but I'm wondering if we can have a comparison suite between conmon and conmon-rs to make sure total memory usage in the system slice doesn't increase a wild amount. maybe cri-o would be a better place for it?
> 
> also: this test is reported as flakey `↺ [FLAKEY TEST - TOOK 3 ATTEMPTS TO PASS] [65.455 seconds]` so that's a risk to what currently exists
> 
> sorry to be pedantic!
> 
> /cat

Yeah sounds good to me, let's do something in CRI-O. :+1:

PR #3197, opened 2026-05-04
Use persistent RPC connection in Go client

saschagrunert:
/kind feature

#### What type of PR is this?

/kind feature

Maintains a single Cap'n Proto RPC connection per `ConmonClient` instead of creating and tearing down a new Unix socket connection for every RPC call. The connection is lazily established on first use and reset on RPC-level errors so the next call reconnects.

None

The `defer free()` pattern and stdout/stderr copy in `ExecSyncContainer` remain unchanged since message lifecycle is independent of connection lifecycle.

```release-note
Use a persistent Cap'n Proto RPC connection per ConmonClient instead of reconnecting on every call.
```

### DIFF
--- a/pkg/client/attach.go
+++ b/pkg/client/attach.go
@@ -104,20 +104,12 @@ func (c *ConmonClient) AttachContainer(ctx context.Context, cfg *AttachConfig) e
 		defer span.End()
 	}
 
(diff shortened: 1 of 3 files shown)


saschagrunert:

PR #3198, opened 2026-05-04
Avoid blocking file removal in Terminal::drop

saschagrunert:
/kind bug

#### What type of PR is this?

/kind bug

Replaces synchronous `std::fs::remove_file()` in `Terminal::drop` with an async `tokio::fs::remove_file()` spawned via `task::spawn`. Since `Drop::drop` is synchronous, the blocking call could stall the Tokio executor thread.

None

```release-note
None
```


saschagrunert:

PR #3264, opened 2026-06-04
perf: journald spawn_blocking, broadcast capacity

saschagrunert:
/kind cleanup

- Wrap journald `journal_print()` calls in per-line `task::spawn_blocking` to prevent blocking tokio worker threads while preserving streaming semantics
- Increase attach broadcast channel capacity from 2 to 16 to reduce message loss with multiple attach clients
- Move `clone()` calls for logger, attach, and message channels inside stream guards to avoid unnecessary allocations when streams are `None`

None

```release-note
None
```


saschagrunert:

PR #3265, opened 2026-06-04
perf: use pidfd for async process exit waiting

saschagrunert:
/kind cleanup

Replace the blocking `waitpid` thread (`task::spawn_blocking`) with `pidfd_open` + `AsyncFd` polling for process exit detection. This eliminates one OS thread per container from tokio's blocking pool, reducing memory usage (thread stacks) and context switching overhead at scale.
Falls back to the blocking `waitpid` approach on kernels older than 5.3 where `pidfd_open` is unavailable.

None

Requires Linux 5.3+ for `pidfd_open`. The fallback path preserves the old behavior exactly.

```release-note
None
```

### DIFF
--- a/conmon-rs/server/src/child_reaper.rs
+++ b/conmon-rs/server/src/child_reaper.rs
@@ -27,7 +27,7 @@ use std::{
 };
 use tokio::{
     fs::{self, File},
-    io::AsyncWriteExt,
+    io::{AsyncWriteExt, unix::AsyncFd},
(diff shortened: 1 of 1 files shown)


saschagrunert:

rphillips:
As a followup, we should see which RHEL releases and kernels have the pidfd patch in it.

saschagrunert:
Codecov blocks, merging.

PR #3363, opened 2026-07-26
server: close file descriptors inherited from the parent

xw19:
### What this PR does

conmon-rs daemonizes and outlives whatever spawned it, so every descriptor it
inherits stays open for the lifetime of the server. If the spawning process was
itself started with extra descriptors open, holding them can stop that process
tree from ever finishing.

This closes everything above stderr right after the daemonizing fork, which is
what `conmon` already does in `close_other_fds()` and `close_all_fds_ge_than(3)`.

### Why

CRI-O's integration suite hits this. bats does not exit until FD 3 is closed by
all of its descendants. The suite backgrounds CRI-O without closing FD 3, CRI-O
spawns conmon-rs, and conmon-rs holds FD 3 for as long as it runs. The suite
then reports every test as passing and hangs until the two hour job timeout.

It only happens on the conmon-rs matrix entry, never conmon — conmon closes
inherited descriptors, conmon-rs did not. Tracked in
https://github.com/cri-o/cri-o/issues/10188, where the full analysis lives.

Observed directly, starting conmon-rs with a pipe on FD 3:

```
=== conmonrs PID 253195 open fds ===
>>> fd 3: /tmp/tmp.PmtQQfFGtB/p
```

### Reproducer

A single bats test that backgrounds conmon-rs and then ends while it is still
running:

```bash
@test "spawn conmonrs and finish the test" {
	D=$(mktemp -d)
	"$CONMONRS_BIN" --runtime /usr/bin/crun --runtime-dir "$D" &>"$D/log" &
	sleep 2
	true
}
```

| conmon-rs build | bats exit | elapsed |
| --- | --- | --- |
| unpatched | 124 | 30s, hung |
| this PR | 0 | 2s |

After the change, FD 3 is conmon-rs's own epoll and no reference to the
inherited pipe remains in its descriptor table.

### Notes for review

- Descriptors conmon-rs needs are either created after this point or received
  later over the fd socket, so nothing above stderr is kept.
- The descriptor used to read `/proc/self/fd` is closed by the time the loop
  runs, so each one is checked with `F_GETFD` before closing to avoid closing an
  unrelated descriptor that reused the number. This runs before tokio starts, so
  nothing else can be opening descriptors concurrently.
- I can switch to `close_range(2)` if you would prefer that over walking
  `/proc/self/fd`; I kept the walk to match what conmon does and to avoid a
  minimum kernel requirement.

### Drafted using Claude

```release-note
None
```

### DIFF
--- a/conmon-rs/server/src/server.rs
+++ b/conmon-rs/server/src/server.rs
@@ -31,7 +31,7 @@ use nix::{
 use opentelemetry::trace::{FutureExt as OpenTelemetryFutureExt, TracerProvider};
 #[cfg(feature = "tracing")]
 use opentelemetry_sdk::trace::SdkTracerProvider;
-use std::{fs::File, io::Write, path::Path, process, str::FromStr, sync::Arc};
+use std::{fs::File, io::Write, os::fd::RawFd, path::Path, process, str::FromStr, sync::Arc};
 use tokio::{
     fs,
     runtime::{Builder, Handle},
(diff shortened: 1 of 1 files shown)


xw19:
@bitoku please take a look

conmon-rs/server/src/server.rs
```diff
@@ -31,7 +31,9 @@ use nix::{
 use opentelemetry::trace::{FutureExt as OpenTelemetryFutureExt, TracerProvider};
 #[cfg(feature = "tracing")]
 use opentelemetry_sdk::trace::SdkTracerProvider;
-use std::{fs::File, io::Write, path::Path, process, str::FromStr, sync::Arc};
+use std::{
+    fs as fs_std, fs::File, io::Write, os::fd::RawFd, path::Path, process, str::FromStr, sync::Arc,
```
saschagrunert: nit: `fs_std` is only used once (line 182). Could inline it as `std::fs::read_dir(...)` at the call site and drop the alias.

conmon-rs/server/src/server.rs
```diff
@@ -147,6 +149,52 @@ impl Server {
...
+    /// keeping.
+    fn close_inherited_fds() {
+        for fd in Self::inherited_fds() {
+            // The descriptor used to enumerate them is already closed by now,
+            // so check before closing to avoid closing an unrelated descriptor
+            // that reused the number. This runs before tokio starts, so nothing
+            // else can be opening descriptors concurrently.
+            // SAFETY: these are descriptors we inherited and never use.
+            if unsafe { libc::fcntl(fd, libc::F_GETFD) } >= 0 {
+                unsafe { libc::close(fd) };
```
saschagrunert: The codebase uses `nix` wrappers elsewhere for syscalls (`fork`, `fcntl` via `FcntlArg`, etc.). Could replace the raw libc calls here to stay consistent and drop the `unsafe` blocks:
```rust
use nix::fcntl::{fcntl, FcntlArg};
use nix::unistd::close;

if fcntl(fd, FcntlArg::F_GETFD).is_ok() {
    let _ = close(fd);
}
```

xw19:
@saschagrunert Thank you for the review.

conmon-rs/server/src/server.rs
```diff
@@ -147,6 +156,58 @@ impl Server {
...
+        for fd in Self::inherited_fds() {
+            // The descriptor used to enumerate them is already closed by now,
+            // so check before closing to avoid closing an unrelated descriptor
+            // that reused the number. This runs before tokio starts, so nothing
+            // else can be opening descriptors concurrently.
+            //
+            // SAFETY: `fd` was listed as open a moment ago and nothing in
+            // between opens descriptors, so borrowing it for the check is
+            // valid.
+            if fcntl(unsafe { BorrowedFd::borrow_raw(fd) }, FcntlArg::F_GETFD).is_ok() {
```
saschagrunert: The SAFETY comment says "nothing in between opens descriptors", but `ReadDir`'s `Drop` *closes* the directory fd when `inherited_fds()` returns. That closed fd number is in the returned `Vec`, so `BorrowedFd::borrow_raw` gets called on a closed fd here.

`BorrowedFd::borrow_raw` requires the fd to remain open for the lifetime of the borrow (`std/src/os/fd/owned.rs:79-80`). The `fcntl` correctly returns `EBADF` and the fd is skipped, so there is no practical issue (it is a zero-cost integer wrapper), but the `unsafe` contract is technically violated.

Simplest fix would be using raw libc for the check, which has no ownership precondition:

```rust
fn close_inherited_fds() {
    for fd in Self::inherited_fds() {
        if unsafe { libc::fcntl(fd, libc::F_GETFD) } != -1 {
            unsafe { libc::close(fd) };
        }
    }
}
```

This also removes the need for the `nix::fcntl` and `nix::unistd::close` imports, and the `BorrowedFd`/`OwnedFd`/`FromRawFd` imports (only `RawFd` is still needed for `inherited_fds`).

conmon-rs/server/src/server.rs
```diff
@@ -147,6 +156,58 @@ impl Server {
...
+            }
+        }
+    }
+
+    /// Every open descriptor above stderr. Split out from `close_inherited_fds`
+    /// so it can be tested without closing the test runner's own descriptors.
+    fn inherited_fds() -> Vec<RawFd> {
+        let mut fds = Vec::new();
+
+        if let Ok(entries) = std::fs::read_dir("/proc/self/fd") {
```
saschagrunert, line 203: This silently does nothing on non-Linux (no `/proc/self/fd`), which means inherited fds would leak on other platforms. Fine since conmon-rs is Linux-only (`prctl`, `epoll`, systemd), but maybe worth a `debug_assert!(cfg!(target_os = "linux"))` or a log line on the `Err` path so it doesn't silently regress if someone tries to port.

conmon-rs/server/src/server.rs
```diff
@@ -164,6 +225,8 @@ impl Server {
             }
         }
 
+        Self::close_inherited_fds();
```
saschagrunert, line 238: nit: this also runs in the `skip_fork` path, which I think is correct (same leak problem applies regardless of forking). Just confirming that's intentional.

saschagrunert (approved):

